PR #435, opened 2015-01-18
Added two new objective c wrappers for libgit2

0x4a616e:
* A patch can directly be copied into a buffer using git_patch_to_buf method
* A tree entry can be looked up by path using git_tree_entry_bypath method. The found object is freed when the GTTreeEntry is deallocated.

### DIFF
--- a/ObjectiveGit/GTDiffPatch.h
+++ b/ObjectiveGit/GTDiffPatch.h
@@ -51,7 +51,7 @@
 - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunkHeaders fileHeaders:(BOOL)includeFileHeaders;
 
 /// Returns the raw patch data.
-- (NSData*)patchData;
+- (NSData *)patchData;
 
 /// Enumerate the hunks contained in the patch.
 ///
--- a/ObjectiveGit/GTDiffPatch.h
+++ b/ObjectiveGit/GTDiffPatch.h
@@ -50,6 +50,9 @@
 /// Returns the raw size of the delta, in bytes.
 - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunkHeaders fileHeaders:(BOOL)includeFileHeaders;
 
+/// Returns the raw patch data as buffer.
+- (NSData*)toBuffer;
+
 /// Enumerate the hunks contained in the patch.
 ///
 /// This enumeration is synchronous, and will block the calling thread while
--- a/ObjectiveGit/GTDiffPatch.h
+++ b/ObjectiveGit/GTDiffPatch.h
@@ -50,8 +50,8 @@
 /// Returns the raw size of the delta, in bytes.
 - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunkHeaders fileHeaders:(BOOL)includeFileHeaders;
 
-/// Returns the raw patch data as buffer.
-- (NSData*)toBuffer;
+/// Returns the raw patch data.
+- (NSData*)patchData;
 
 /// Enumerate the hunks contained in the patch.
 ///
--- a/ObjectiveGit/GTDiffPatch.m
+++ b/ObjectiveGit/GTDiffPatch.m
@@ -63,7 +63,7 @@
 	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
 	git_patch_to_buf(&buf, self.git_patch);
 
-	NSData* buffer = [[NSData alloc] initWithBytes:buf.ptr length:buf.size];
+	NSData *buffer = [[NSData alloc] initWithBytes:buf.ptr length:buf.size];
 	git_buf_free(&buf);
 	
 	return buffer;
--- a/ObjectiveGit/GTDiffPatch.m
+++ b/ObjectiveGit/GTDiffPatch.m
@@ -59,6 +59,16 @@
 	return git_patch_size(self.git_patch, includeContext, includeHunkHeaders, includeFileHeaders);
 }
 
+- (NSData*)toBuffer {
+	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
+	git_patch_to_buf(&buf, self.git_patch);
+
+	NSData* buffer = [[NSData alloc] initWithBytes:buf.ptr length:buf.size];
+	git_buf_free(&buf);
+	
+	return buffer;
+}
+
 #pragma mark Hunks
 
 - (BOOL)enumerateHunksUsingBlock:(void (^)(GTDiffHunk *hunk, BOOL *stop))block {
--- a/ObjectiveGit/GTDiffPatch.m
+++ b/ObjectiveGit/GTDiffPatch.m
@@ -59,7 +59,7 @@
 	return git_patch_size(self.git_patch, includeContext, includeHunkHeaders, includeFileHeaders);
 }
 
-- (NSData*)patchData {
+- (NSData *)patchData {
 	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
 	git_patch_to_buf(&buf, self.git_patch);
 
--- a/ObjectiveGit/GTDiffPatch.m
+++ b/ObjectiveGit/GTDiffPatch.m
@@ -59,7 +59,7 @@
 	return git_patch_size(self.git_patch, includeContext, includeHunkHeaders, includeFileHeaders);
 }
 
-- (NSData*)toBuffer {
+- (NSData*)patchData {
 	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
 	git_patch_to_buf(&buf, self.git_patch);
 
--- a/ObjectiveGit/GTTree.h
+++ b/ObjectiveGit/GTTree.h
@@ -64,6 +64,8 @@ typedef NS_ENUM(NSInteger, GTTreeEnumerationOptions) {
 /// returns a GTTreeEntry or nil if there is nothing with the specified name
 - (GTTreeEntry *)entryWithName:(NSString *)name;
 
+- (GTTreeEntry*) treeEntryByPath:(NSString*) path error:(NSError**)error;
+
 /// Enumerates the contents of the tree
 ///
 /// options -  One of `GTTreeEnumerationOptionPre` (for pre-order walks) or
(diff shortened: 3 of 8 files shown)


ObjectiveGit/GTDiffPatch.h
```diff
@@ -50,6 +50,9 @@
 /// Returns the raw size of the delta, in bytes.
 - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunkHeaders fileHeaders:(BOOL)includeFileHeaders;
 
+/// Returns the raw patch data as buffer.
+- (NSData*)toBuffer;
```
tiennou: `patchData` feels like a more "conventional" name.

ObjectiveGit/GTDiffPatch.m
```diff
@@ -59,6 +59,16 @@ - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunk
 	return git_patch_size(self.git_patch, includeContext, includeHunkHeaders, includeFileHeaders);
 }
 
+- (NSData*)toBuffer {
+	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
+	git_patch_to_buf(&buf, self.git_patch);
+
+	NSData* buffer = [[NSData alloc] initWithBytes:buf.ptr length:buf.size];
```
tiennou: You *could* prevent a copying using `-initWithBytesNoCopy:length:deallocator:`, but that's 10.9+ and I don't think we're that high yet.

ObjectiveGit/GTTree.h
```diff
@@ -64,6 +64,8 @@ typedef NS_ENUM(NSInteger, GTTreeEnumerationOptions) {
 /// returns a GTTreeEntry or nil if there is nothing with the specified name
 - (GTTreeEntry *)entryWithName:(NSString *)name;
 
+- (GTTreeEntry*) treeEntryByPath:(NSString*) path error:(NSError**)error;
```
tiennou: Style: no space before method name.

ObjectiveGit/GTTree.m
```diff
@@ -68,6 +68,17 @@ - (GTTreeEntry *)entryWithName:(NSString *)name {
 	return [self createEntryWithEntry:git_tree_entry_byname(self.git_tree, name.UTF8String)];
 }
 
+- (GTTreeEntry*) treeEntryByPath:(NSString*)path error:(NSError**)error {
```
tiennou: Maybe `-entryWithPath:` would be a better name, matching the ones above ?

ObjectiveGitFramework.xcodeproj/project.pbxproj
```diff
@@ -412,6 +412,7 @@
 		30DCBA7017B4791A009B0EBD /* NSArray+StringArray.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = "NSArray+StringArray.m"; sourceTree = "<group>"; };
 		30FDC07D16835A8100654BF0 /* GTDiffLine.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = GTDiffLine.h; sourceTree = "<group>"; };
 		30FDC07E16835A8100654BF0 /* GTDiffLine.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = GTDiffLine.m; sourceTree = "<group>"; };
+		3D6123BB1A643296008F831A /* libgit2.0.21.2.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = libgit2.0.21.2.dylib; path = ../../../../../usr/local/Cellar/libgit2/0.21.2/lib/libgit2.0.21.2.dylib; sourceTree = "<group>"; };
```
tiennou: I don't think adding this to the project file is necessary. Could you remove those changes ?

ObjectiveGit/GTTreeEntry.m
```diff
@@ -39,6 +39,7 @@
 
 @interface GTTreeEntry ()
 @property (nonatomic, assign, readonly) const git_tree_entry *git_tree_entry;
+@property (nonatomic, assign, readonly) git_tree_entry* entry_to_free;
```
tiennou: I find this a little fishy, so I'll let the people in the know give their opinion ;-).

I *think* the `const` on the line above should be dropped so `GTTreeEntry` assumes ownership of the underlying `git_tree_entry`, then this duplication becomes unnecessary.

0x4a616e: I thougt about that, but that would cause interface changes that might break other projects.
Furthermore, the existing modifier is currently needed since methods like "git_tree_entry_byname" return "const git_tree_entry *". One solution might be to copy those tree entries before adding them to GTTreeEntry.

tiennou: After a closer look, [git_tree_entry_byname](https://libgit2.github.com/libgit2/#HEAD/group/tree/git_tree_entry_byname) returns a pointer that is managed internally. As such, there is no need to call `git_tree_entry_free` on it, so the code in `-[GTTree treeEntryByPath:]` above can be changed to the simpler version `-entryWithName:` uses.

joshaber:
The different ownership semantics make me wonder if we should just `git_tree_entry_dup` all entries so that we're consistent.

0x4a616e:
That should solve this, I'll change it

0x4a616e:
Creating a new pull request incorporating your feedback

ObjectiveGit/GTDiffPatch.h
```diff
@@ -50,6 +50,9 @@
 /// Returns the raw size of the delta, in bytes.
 - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunkHeaders fileHeaders:(BOOL)includeFileHeaders;
 
+/// Returns the raw patch data.
+- (NSData*)patchData;
```
joshaber: Style note (here and elsewhere): space between the type and the asterisk.

joshaber:
I'd be interested in seeing what others think about this. My fear is that the current interface still makes it too easy to mess up the ownership of the tree entry.

ObjectiveGit/GTTree.h
```diff
@@ -64,6 +64,13 @@ typedef NS_ENUM(NSInteger, GTTreeEnumerationOptions) {
 /// returns a GTTreeEntry or nil if there is nothing with the specified name
 - (GTTreeEntry *)entryWithName:(NSString *)name;
 
+/// Get a entry by path
```
tiennou: Nitpick: a/an.

jspahrsummers:
I agree with @joshaber that we should stick to _one_ ownership semantic for this class. It's simpler, even if slightly suboptimal for some use cases.

ObjectiveGit/GTTree.m
```diff
@@ -56,16 +56,27 @@ - (NSUInteger)entryCount {
...
 - (GTTreeEntry *)entryWithName:(NSString *)name {
-	return [self createEntryWithEntry:git_tree_entry_byname(self.git_tree, name.UTF8String)];
+	return [self createEntryWithCopyOfEntry:git_tree_entry_byname(self.git_tree, name.UTF8String)];
+}
+
+- (GTTreeEntry *)entryWithPath:(NSString *)path error:(NSError **)error {
+	git_tree_entry *entry = NULL;
+	int gitError = git_tree_entry_bypath(&entry, self.git_tree, path.UTF8String);
+	if (error != GIT_OK) {
+		*error = [NSError git_errorFor:gitError description:@"Failed to get tree ntry %@", path];
```
tiennou: If `NULL` is passed you'll crash.

tiennou:
Agreed about the ownership thing. Though the fact is that *some* functions return externally-owned tree entries while others don't.

I propose `-[GTTreeEntry initWithEntry:parentTree:freeWhenDone:error:]` to be added and make the default method to assume `shouldFree` (changing semantics for existing users though). And externally-owned users would pass `NO`, doing the right thing when the entry goes away ?

ObjectiveGit/GTTree.h
0x4a616e: That was actually a copy-paste error ;-) I'll fix the other occurrences as well.

joshaber:
>Though the fact is that some functions return externally-owned tree entries while others don't.

But if we call `git_tree_entry_dup` on those entries, we'll always end up with an objective-git-owned entry, right?

0x4a616e:
What do you think about switching the interface back to `const git_tree_entry*` and always creating a copy of the entry internally within GTTreeEntry. This way, the object would always be solely owned and freed by GTTreeEntry.

joshaber:
Sounds good :+1:

tiennou:
My goal with the "private" method was that if there's someone out there who uses libgit2 directly and has to wrap one of the externally-owned entries itself, they it will have to duplicate what we're doing internally.

But yeah, always duping is good, as it actually will prevent some future weird crashes where the underlying owner gets freed (thus freeing entries) while there's still some `GTTreeEntry` in e.g. an array.

0x4a616e:
In the latest version, the git_tree_entry is always duplicated. What do you think? :)

ObjectiveGit/GTDiffPatch.m
```diff
@@ -59,6 +59,16 @@ - (NSUInteger)sizeWithContext:(BOOL)includeContext hunkHeaders:(BOOL)includeHunk
 	return git_patch_size(self.git_patch, includeContext, includeHunkHeaders, includeFileHeaders);
 }
 
+- (NSData *)patchData {
+	git_buf buf = GIT_BUF_INIT_CONST(0, NULL);
+	git_patch_to_buf(&buf, self.git_patch);
+
+	NSData* buffer = [[NSData alloc] initWithBytes:buf.ptr length:buf.size];
```
joshaber: Style: `NSData *buffer`

ObjectiveGit/GTTreeEntry.m
```diff
@@ -64,19 +64,33 @@ - (BOOL)isEqualToEntry:(GTTreeEntry *)treeEntry {
...
+- (instancetype)initWithEntry:(const git_tree_entry *)theEntry parentTree:(GTTree *)parent error:(NSError **)error {
 	NSParameterAssert(theEntry != NULL);
-	if((self = [super init])) {
-		_git_tree_entry = theEntry;
-		_tree = parent;
+	
+	self = [super init];
+	if (self == nil) return nil;
+	
+	git_tree_entry* copyOfEntry = nil;
```
joshaber: Style: `git_tree_entry *copyOfEntry`

joshaber:
Awesome, thanks! I didn't realize the changes were made.

Just a couple more style notes.

0x4a616e:
Style changed

joshaber:
:metal:

PR #437, opened 2015-01-20
Update libgit2

joshaber:
Unfortunately libgit2’s push API has changed significantly so I had to remove our wrapper for now.

/cc @phatblat


joshaber:
Missed some things.

joshaber:
Ready now :dancer:

phatblat:
:frowning: I'll take a look at the changes to the push API.

Issue #436, opened 2015-01-19
Including library in Swift only project is failing


pietbrauer:
I tried installing this framework using Carthage but when I try to import it into my Swift project is tells me `"git2.h" not found`.

Also I feel that the Installation Instructions are a bit outdated for iOS, so I can't get it working.

jspahrsummers:
Ah, sorry about the outdated instructions! It should basically be the same [as for OS X](https://github.com/libgit2/objective-git/blob/d3ecf14a75ffbbaa813c5cd6c13ecdc6916546dd/README.md#importing-objectivegit-on-os-x) now.

pietbrauer:
Will try! If it works, I will make a PR and update the instructions.

pietbrauer:
Well, this also failed with some weird errors.
I tried using Carthage again and this also failed (see build log). It also took a bit (15 min) and carthage wasn't really providing much output, will open another issue if you feel this is an issue too.
```bash
*** Fetching objective-git
*** Checking out objective-git at "0.5.1"
*** xcodebuild output can be found in /var/folders/9v/9z2qkhhx57zbdn9876mq__497d7w_y/T/carthage-xcodebuild.4Q12iP.log
*** Building scheme "ObjectiveGit iOS" in ObjectiveGitFramework.xcworkspace
2015-01-19 21:33:47.457 xcodebuild[14359:1173548] [MT] PluginLoading: Required plug-in compatibility UUID C4A681B0-4A26-480E-93EC-1218098B9AA0 for plug-in at path '~/Library/Application Support/Developer/Shared/Xcode/Plug-ins/SnapshotDiffs.xcplugin' not present in DVTPlugInCompatibilityUUIDs
** BUILD FAILED **


The following commands produced analyzer issues:
	AnalyzeShallow ObjectiveGit/GTDiffLine.m
	AnalyzeShallow ObjectiveGit/GTTreeBuilder.m
	AnalyzeShallow ObjectiveGit/GTIndexEntry.m
	AnalyzeShallow ObjectiveGit/GTEnumerator.m
	AnalyzeShallow ObjectiveGit/GTConfiguration.m
	AnalyzeShallow ObjectiveGit/GTRepository+Attributes.m
	AnalyzeShallow ObjectiveGit/ObjectiveGit.m
	AnalyzeShallow ObjectiveGit/GTCommit.m
	AnalyzeShallow ObjectiveGit/GTTree.m
[...]

PR #438, opened 2015-01-21
Add modulemap file to iOS target

pietbrauer:
As discussed in #436 adding a module map file in iOS is required as it is in OS X.


mdiep:
I can't build the iOS target with or without these changes, but this looks legit to me.

pietbrauer:
There is no CI for iOS right?

jspahrsummers:
CI only works for collaborators, but yeah—there are no iOS tests, partly due to the ridiculous build times.

jspahrsummers:
Anyways, I agree that this is legit. :+1:

pietbrauer:
I see. Thanks!

PR #439, opened 2015-02-04
Add ability to cancel the remote part of the clone operation

alehed:
This makes the transferProgressBlock look more like its libgit equivalent
the git_transfer_progress_cb.


alehed:
I don't know whether it is worth introducing a breaking change for such a small 
feature but in my opinion objective-git should have allowed this in the first place. 
Feel free to take away the comment, I just thought it would be good documenting 
the change somewhere.

joshaber:
I think it's a great idea to be able to cancel the transfer, but I think it'd be more idiomatic in Objective-C for the block to set a stop argument. E.g.,

```objc
transferProgressBlock:(void (^)(const git_transfer_progress *, BOOL *stop))transferProgressBlock
```

phatblat:
+1 for `*stop`

alehed:
I think this should do the trick

joshaber:
Awesome, thanks @alehed!

Issue #442, opened 2015-02-16
0.6 won't build


pietbrauer:
When trying to build the 0.6 release I always get the following error when trying to execute `script/update_libgit2_ios` although `libssh2` and `openssl` built successful.

[jspahrsummers seems to have the same problem](https://twitter.com/jspahrsummers/status/566824855794761728)

```
[ 90%] Building C object CMakeFiles/git2.dir/src/transports/smart_protocol.c.o
[ 91%] Building C object CMakeFiles/git2.dir/src/transports/ssh.c.o
/Users/piet.brauer/Documents/Projekte/iOS/objective-git-build/Carthage/Checkouts/objective-git/External/libgit2/src/transports/ssh.c:9:10: fatal error: 'libssh2.h' file not found
#include <libssh2.h>
         ^
1 error generated.
make[2]: *** [CMakeFiles/git2.dir/src/transports/ssh.c.o] Error 1
make[1]: *** [CMakeFiles/git2.dir/all] Error 2
make: *** [all] Error 2
```

phatblat:
Same here. I was able to get it to build by defining LIBSSH2_INCLUDE_DIR. 

See https://github.com/phatblat/objective-git/commit/b817e13e4c29bea355044e316d9214fbc955ac8f

I had to make a similar change to get the Mac build to work but haven't included these changes in a pull request yet.

pietbrauer:
@phatblat Nice work! I so look forward to have prebuilt frameworks with Carthage!!

pietbrauer:
@phatblat Worked for me.

jspahrsummers:
Could one of you please open a PR for this? :pray:

PR #443, opened 2015-02-16
Fix libgit2 build scripts

phatblat:
I had some trouble getting libgit2 to build in 0.5 and 0.6, getting errors about `'libssh2.h' file not found`. I wasn't sure if it was my setup (been on Xcode and Yosemite betas a lot lately), but by modifying the update_libgit2 scripts and defining `LIBSSH2_INCLUDE_DIR` it builds consistently.

Addresses #442

### DIFF
--- a/script/update_libgit2
+++ b/script/update_libgit2
@@ -22,7 +22,11 @@ fi
 mkdir build
 cd build
 
-cmake -DBUILD_SHARED_LIBS:BOOL=OFF -DBUILD_CLAR:BOOL=OFF -DTHREADSAFE:BOOL=ON ..
+cmake -DBUILD_SHARED_LIBS:BOOL=OFF \
+	-DLIBSSH2_INCLUDE_DIR:PATH=../../External/libssh2-ios/include/libssh2/ \
+	-DBUILD_CLAR:BOOL=OFF \
+	-DTHREADSAFE:BOOL=ON \
+	..
 cmake --build .
 
 product="libgit2.a"
--- a/script/update_libgit2
+++ b/script/update_libgit2
@@ -22,11 +22,7 @@ fi
 mkdir build
 cd build
 
-cmake -DBUILD_SHARED_LIBS:BOOL=OFF \
-	-DLIBSSH2_INCLUDE_DIR:PATH=../../External/libssh2-ios/include/libssh2/ \
-	-DBUILD_CLAR:BOOL=OFF \
-	-DTHREADSAFE:BOOL=ON \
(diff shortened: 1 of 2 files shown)


script/update_libgit2
```diff
@@ -22,7 +22,11 @@ fi
 mkdir build
 cd build
 
-cmake -DBUILD_SHARED_LIBS:BOOL=OFF -DBUILD_CLAR:BOOL=OFF -DTHREADSAFE:BOOL=ON ..
+cmake -DBUILD_SHARED_LIBS:BOOL=OFF \
+	-DLIBSSH2_INCLUDE_DIR:PATH=../../External/libssh2-ios/include/libssh2/ \
```
jspahrsummers: This won't be correct for OS X builds, which use the system libssh2 (IIRC).

jspahrsummers:
Thanks! The iOS changes definitely look good, but I'm a bit unsure about the OS X path.

phatblat:
Went ahead and reverted the change to the Mac build script. When I run `script/update_libgit2` from the command line it builds fine.

It's only when I build the libgit2 target in Xcode that it fails.

```
[ 88%] Building C object CMakeFiles/git2.dir/src/transports/smart.c.o
[ 89%] Building C object CMakeFiles/git2.dir/src/transports/smart_pkt.c.o
[ 90%] Building C object CMakeFiles/git2.dir/src/transports/smart_protocol.c.o
[ 91%] Building C object CMakeFiles/git2.dir/src/transports/ssh.c.o
/Users/phatblat/dev/ios/APP_NAME/Carthage/Checkouts/objective-git/External/libgit2/src/transports/ssh.c:9:10: fatal error: 'libssh2.h' file not found
#include <libssh2.h>
         ^
1 error generated.
make[2]: *** [CMakeFiles/git2.dir/src/transports/ssh.c.o] Error 1
make[1]: *** [CMakeFiles/git2.dir/all] Error 2
make: *** [all] Error 2
```

pietbrauer:
For me both (Mac and iOS) built when I checked out @phatblat branch. Just using the iOS changes didn't solve the building issues.

jspahrsummers:
Sorry, to clear: I think `LIBSSH2_INCLUDE_DIR` is important to pass to the Mac target, I just don't think it should point into the iOS build of libssh2.

phatblat:
`/usr/local/include` works

jspahrsummers:
Awesome, thanks for tackling this! :sparkles:

PR #445, opened 2015-03-03
Safe signature

joshaber:
Create a signature using default values if libgit2’s validation fails.

### DIFF
--- a/ObjectiveGit/GTRepository.m
+++ b/ObjectiveGit/GTRepository.m
@@ -718,24 +718,33 @@ static int submoduleEnumerationCallback(git_submodule *git_submodule, const char
 
 #pragma mark User
 
++ (NSString *)defaultUserName {
+	NSString *name = NSFullUserName();
+	if (name.length == 0) name = NSUserName();
+	if (name.length == 0) name = @"nobody";
+	return name;
+}
+
++ (NSString *)defaultEmail {
+	NSString *username = NSUserName();
+	if (username.length == 0) username = @"nobody";
+	NSString *domain = NSProcessInfo.processInfo.hostName ?: @"nowhere.local";
+	return [NSString stringWithFormat:@"%@@%@", username, domain];
+}
+
 - (GTSignature *)userSignatureForNow {
 	GTConfiguration *configuration = [self configurationWithError:NULL];
 	NSString *name = [configuration stringForKey:@"user.name"];
-	if (name.length == 0) {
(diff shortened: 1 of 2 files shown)


ObjectiveGit/GTRepository.m
```diff
@@ -718,24 +718,33 @@ - (GTSubmodule *)submoduleWithName:(NSString *)name error:(NSError **)error {
...
+	if (name.length == 0) name = self.class.defaultUserName;
 
 	NSString *email = [configuration stringForKey:@"user.email"];
-	if (email == nil) {
-		NSString *username = NSUserName();
-		if (username.length == 0) username = @"nobody";
-		NSString *domain = NSProcessInfo.processInfo.hostName ?: @"nowhere.local";
-		email = [NSString stringWithFormat:@"%@@%@", username, domain];
-	}
+	if (email == nil) email = self.class.defaultEmail;
```
jspahrsummers: Why shouldn't this be a length check either?

jspahrsummers:
Can we unart test this too?

ObjectiveGit/GTRepository.m
joshaber: Libgit2 used to be fine with empty emails: https://github.com/libgit2/libgit2/commit/76e3c43fb99eb75cf33affb1a93260af4d472da5

Apparently not anymore :cry:

joshaber:
I went back and forth on whether we should test for `<` and `>` in the email explicitly. That'd let us be more useful, but it'd be duplicating libgit2's test. Thoughts?

jspahrsummers:
> That'd let us be more useful

What do you mean? I was thinking we could just test that invalid strings result in the default.

joshaber:
Ah, sorry, I meant that if we detect that the email address has `<` or `>` in it (which means libgit2 will reject it), we could still try using the configured user name. As is, libgit2 would reject the whole signature and we'd fallback to the default username and email.

jspahrsummers:
If there's a libgit2 API to check validity without actually instantiating a signature, we could test that with the given username and email separately, then both together.

If that sort of thing doesn't exist, I think this implementation is fine.

joshaber:
>If there's a libgit2 API to check validity without actually instantiating a signature, we could test that with the given username and email separately, then both together.

Sadly no.

:fallen_leaf:

jspahrsummers:
:leaves:

PR #450, opened 2015-03-21
Support OIDs for ahead/behind count and enumeration

jspahrsummers:
Also changed `GTBranch` and `GTReference` to expose OIDs, not SHAs.


robrix:
Looks great :+1:

Issue #448, opened 2015-03-21
Add support for Swift 1.2 / Xcode 6.3


phatblat:
- Update Quick & Nimble
- New Objective-C nonnull and nullable attributes

PR #452, opened 2015-03-27
Support for Swift 1.2 (Spec dependencies) and new Objective-C nullability qualifiers

phatblat:
Updated Quick (0.3.1) and Nimble (0.4.2), which are both now using Swift 1.2. :warning: These changes require Xcode 6.3.

One of the nice features for Objective-C in 6.3 is new [nullability qualifiers](https://developer.apple.com/swift/blog/?id=25). I did an audit of the API, marking parameters, return types and properties as `nullable` where documented, but also verifying behavior in the implementation. I did not go all the way into the libgit2 implementation but rather erred on the `nonnull` side in most cases.

IMO, the `NS_ASSUME_NONNULL_BEGIN` / `END` macros have more signal to noise than ` #pragma clang assume_nonnull begin` / `end`, but either of them are better than sprinkling `nonnull` on everything. Plus, this mirrors what's appearing on the Swift side: (assumed) `nonnull` become non-optional and `nullable` become optional. Much safer than forced unwrap on everything.

Also added some missing doc comments and switched initializer return types from `id` to `instancetype`.

Tested with [Xcode 6.3](https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW306)

Addresses #448

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
 github "jspahrsummers/xcconfigs" >= 0.7.1
-github "Quick/Quick" ~> 0.2
-github "Quick/Nimble" ~> 0.2
+github "Quick/Quick" ~> 0.3
+github "Quick/Nimble" ~> 0.4
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,3 +1,3 @@
-github "Quick/Nimble" "v0.4.1"
-github "Quick/Quick" "v0.3.0"
+github "Quick/Nimble" "v0.4.2"
+github "Quick/Quick" "v0.3.1"
 github "jspahrsummers/xcconfigs" "0.7.2"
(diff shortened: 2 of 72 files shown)


joshaber:
Wow, this is great! That's for taking that on!

This'll obviously need to wait until 6.3 is public, but otherwise looks :sparkles:

phatblat:
Yep, completely understood. CI will also need to be switched to Xcode 6.3. Where is the CI configured for this project? Is it on an internal server or up on Travis?

joshaber:
Yup, exactly :+1: 

CI is an internal server but run only on non-fork branches.

phatblat:
Updated with master. This is ready for review.

joshaber:
:metal: 

We're still working on getting our CI server updated. Then this should be :sparkles:

jspahrsummers:
I think we could merge this before CI is ready. :open_hands:

phatblat:
Updated with new API from #463.

joshaber:
You're a hero :sparkling_heart: We're (hopefully) really close to being able to merge this.

joshaber:
Again, thanks so much for doing this @phatblat :sparkling_heart:. Let's do it :metal:

phatblat:
:tada:

PR #455, opened 2015-04-08
Fix entryWithPath:error: on GTTree

sagmor:
`entryWithPath:error:` was failing when error was not `nil`.

Fixed the typo and added some specs for that functionality.


joshaber:
Nice catch! Thanks! :metal:

PR #458, opened 2015-04-09
Add/Fetch NSData from index

sagmor:
Hi,

I added support for libgit2 `git_index_add_frombuffer` function on `GTIndex` with the following methods:

```
// - (NSData *)dataWithName:(NSString *)name error:(NSError **)error; (removed)
- (BOOL)addData:(NSData *)data withName:(NSString *)name error:(NSError **)error;
```

I still have to add some documentation and need to make it work with an `inMemoryIndexWithRepository:error:`

Also I had to update `External/libgit2` to get the method, not sure if there's an issue with this.
Right now GTIndexSpec is passing but have failing specs at `GTConfigurationSpec` and `GTRemoteSpec`

So, I'm still working on it but I'm pushing this in case you find something fishy (specially with the updated dependency and the two failing specs)

### DIFF
--- a/ObjectiveGit/GTIndex.h
+++ b/ObjectiveGit/GTIndex.h
@@ -114,10 +114,6 @@
 /// Returns a new GTIndexEntry, or nil if an error occurred.
 - (GTIndexEntry *)entryWithName:(NSString *)name error:(NSError **)error;
 
-///TODO: Document
-/// Get the 
-- (NSData *)dataWithName:(NSString *)name error:(NSError **)error;
-
 /// Add an entry to the index.
 ///
 /// Note that this *cannot* add submodules. See -[GTSubmodule addToIndex:].
--- a/ObjectiveGit/GTIndex.h
+++ b/ObjectiveGit/GTIndex.h
@@ -114,6 +114,9 @@
 /// Returns a new GTIndexEntry, or nil if an error occurred.
 - (GTIndexEntry *)entryWithName:(NSString *)name error:(NSError **)error;
 
+///TODO: Document
+- (NSData *)dataWithName:(NSString *)name error:(NSError **)error;
+
 /// Add an entry to the index.
 ///
 /// Note that this *cannot* add submodules. See -[GTSubmodule addToIndex:].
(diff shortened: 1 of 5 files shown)


sagmor:
I think I will move the data fetching to `GTIndexEntry` to make it work more like `GTTreeEntry` for API consistency.

sagmor:
Now this is working better.

The only problem I'm having is that this doesn't work with in memory indexes because `git_index_add_frombuffer` fetches the repository from the struct's `rc.owner` property which looks like it can't be set without touching libgit2.

Let me know if you see any issue with this PR :+1:

ObjectiveGit/GTIndex.h
```diff
@@ -135,6 +135,14 @@
 /// Returns YES if successful, NO otherwise.
 - (BOOL)addFile:(NSString *)file error:(NSError **)error;
 
+/// Add an entry (with the provided data and name) to the index.
+/// Will fail if the receiver's repository is nil.
+///
+/// data  - The content of the entry to add
```
joshaber: Looks like these can't be nil according to the implementation. Could we document that?

ObjectiveGit/GTIndexEntry.m
```diff
@@ -86,4 +96,44 @@ - (GTIndexEntryStatus)status {
...
+@end
+
+@implementation GTObject (GTIndexEntry)
+
++ (instancetype)objectWithIndexEntry:(GTIndexEntry *)indexEntry error:(NSError **)error {
+	return [[self alloc] initWithIndexEntry:indexEntry error:error];
+}
+
+- (instancetype)initWithIndexEntry:(GTIndexEntry *)indexEntry error:(NSError **)error {
+
```
joshaber: Don't need this newline.

ObjectiveGit/GTIndexEntry.m
```diff
@@ -86,4 +96,44 @@ - (GTIndexEntryStatus)status {
...
+	git_object *obj;
+	int gitError = git_object_lookup(&obj, indexEntry.repository.git_repository, indexEntry.OID.git_oid, (git_otype)GTObjectTypeAny);
+
+	if (gitError < GIT_OK) {
+		if (error != NULL) {
+    		*error = [NSError git_errorFor:gitError description:@"Failed to get object for index entry."];
```
joshaber: This indentation's a little funny.

joshaber:
Awesome, thanks! :beers: 

>The only problem I'm having is that this doesn't work with in memory indexes because git_index_add_frombuffer fetches the repository from the struct's rc.owner property which looks like it can't be set without touching libgit2.

No worries. Just a couple other notes.

sagmor:
There it is :smile:

I addressed those notes and a couple more I found around

joshaber:
:metal: 

Thanks!

sagmor:
Cool :+1:

PR #460, opened 2015-04-13
API consistency sugestions

sagmor:
Here I'm proposing two small API changes.
I recently started working with this library and some names felt misleading to me at first. So I presume they might be misleading to others:

First, `GTIndex` operations are performed with paths but that's not clear from the API so I changed all references like `entryWithName:` to `entryWithPath:`

Then on `GTBranch`, `targetCommitAndReturnError:` felt strange ¿Is it returning an error or a commit? so I renamed it to `targetCommitWithError:` 

I kept the old methods with a deprecation warning.

:warning: this PR is based on #452 so merging this will merge that PR too, needed that because XCode 6.3 was released.

### DIFF
--- a/ObjectiveGit/GTBranch.h
+++ b/ObjectiveGit/GTBranch.h
@@ -75,7 +75,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// error(out) - will be filled if an error occurs
 ///
 /// returns a GTCommit object or nil if an error occurred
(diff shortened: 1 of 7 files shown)


joshaber:
Those changes sound fine to me. Obviously we'll have to wait on #452 to land first :cookie:

sagmor:
:tada:

PR #461, opened 2015-04-28
Update libssh2 submodule url

phatblat:
Moved to github as of 2015-03-23. There have been some commits but no official release since the move, so no need to update the commit yet. The biggest gain from the switch is no more invalid certificates.

```
fatal: unable to access 'https://git.libssh2.org/libssh2.git/': SSL certificate problem: Invalid certificate chain
Clone of 'git://git.libssh2.org/libssh2.git' into submodule path 'External/libssh2' failed
```

old: https://git.libssh2.org/?p=libssh2.git;a=shortlog;js=1
new: https://github.com/libssh2/libssh2/commits/master

announcement: http://www.libssh2.org/mail/libssh2-devel-archive-2015-03/0029.shtml

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/openssl/openssl.git
 [submodule "libssh2"]
 	path = External/libssh2
-	url = git://git.libssh2.org/libssh2.git
+	url = https://github.com/libssh2/libssh2.git
 [submodule "Carthage.checkout/Nimble"]
 	path = Carthage/Checkouts/Nimble
 	url = https://github.com/Quick/Nimble.git


joshaber:
Oh cool! Even better, it looks like once we bump libgit2 we won't need OpenSSL at all: https://github.com/libgit2/libgit2/pull/2997

mdiep:
I think it's still needed for iOS?

phatblat:
Oh nice! Does the elimination of OpenSSL remove the need for libssh2?

joshaber:
I :doughnut: know the answer to either of those questions :smile:

phatblat:
libssh2 depends on OpenSSL, mostly for the crypto stuff, I believe. Because of this, ObjectiveGit still has a transitive dependency on OpenSSL, which means the libgit2 change won't improve the iOS build times :disappointed:  The OS X build gets pre-built libssh2 and openssl from brew.

I take @carlosmn's comment to mean there _may_ eventually be a change that removes the libssh2 dependency.

> There is still of course libssh2 which will still link against openssl, so we're not quite free of threading nonsense yet, but someone was trying to upstream a SecureTransport version.

carlosmn:
libssh2 is an optional dependency, you can build without it you want; but libssh2 may lose its OpenSSL dependency, which would make it easier to write threaded programs.

phatblat:
Found some work on a libssh2 fork adding support for SecureTransport https://github.com/keithduncan/libssh2/pull/2

PR #462, opened 2015-06-02
Avoid calling tag enumerator block if lookup returns nil tag

dleehr:
Tag lookup may fail in `GTRepositoryForeachTagCallback`, resulting in a nil value for `tag`. This change skips the block if `tag` is nil.

Otherwise, a nil tag may be passed to the block. In the case of `allTagsWithError:`, the nil is added to an array, causing an exception or crash.

### DIFF
--- a/ObjectiveGit/GTRepository.m
+++ b/ObjectiveGit/GTRepository.m
@@ -457,7 +457,7 @@ static int GTRepositoryForeachTagCallback(const char *name, git_oid *oid, void *
 	GTTag *tag = (GTTag *)[info->myself lookUpObjectByGitOid:oid objectType:GTObjectTypeTag error:NULL];
 
 	BOOL stop = NO;
-	if (tag) {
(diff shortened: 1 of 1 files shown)


ObjectiveGit/GTRepository.m
```diff
@@ -457,7 +457,9 @@ static int GTRepositoryForeachTagCallback(const char *name, git_oid *oid, void *
 	GTTag *tag = (GTTag *)[info->myself lookUpObjectByGitOid:oid objectType:GTObjectTypeTag error:NULL];
 
 	BOOL stop = NO;
-	info->block(tag, &stop);
+	if (tag) {
```
joshaber: Style: do an explicit comparison `!= nil`

joshaber:
Thanks! Just one note.

dleehr:
Fixed, thanks for the tip.

joshaber:
:boom:

PR #466, opened 2015-06-22
[WIP] Bump libgit2

joshaber:
This updates libgit2.

Unfortunately it looks like all the remote APIs changed again :cry:

I hate to burn them down again, but it’s also bothersome that the APIs keep changing.

Anyone have thoughts about how we should address this?

/cc @phatblat

### DIFF
--- a/ObjectiveGit/GTDiffDelta.m
+++ b/ObjectiveGit/GTDiffDelta.m
@@ -73,7 +73,7 @@ static int GTDiffDeltaCallback(const git_diff_delta *delta, float progress, void
 	__block git_diff_delta diffDelta;
 
 	int returnValue = [GTDiff handleParsedOptionsDictionary:options usingBlock:^(git_diff_options *optionsStruct) {
-		return git_diff_blobs(oldBlob.git_blob, oldBlobPath.UTF8String, newBlob.git_blob, newBlobPath.UTF8String, optionsStruct, &GTDiffDeltaCallback, NULL, NULL, &diffDelta);
+		return git_diff_blobs(oldBlob.git_blob, oldBlobPath.UTF8String, newBlob.git_blob, newBlobPath.UTF8String, optionsStruct, &GTDiffDeltaCallback, NULL, NULL, NULL, &diffDelta);
 	}];
 
 	if (returnValue != GIT_OK) {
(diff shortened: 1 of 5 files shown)


mdiep:
If we're going to `#ifdef` out the code, we should include the declaration in that so it's obvious to anyone who's using them.

joshaber:
Yup. :+1: I was just doing that so I could get the rest of it building to make sure I didn't miss any API changes.

phatblat:
Who knew push would be so :fire:? I'll take a look at the API changes and see about updating these disabled calls.

phatblat:
Made things happy with the API changes, but not 100% certain everything is correct yet as there are a handful of failing tests.

phatblat:
Down to one broken test which is failing because a newly created remote is having the expected `git://` scheme changed to `https://`. Does libgit2 no longer support the git protocol?

pietbrauer:
What's the usual update process for `libgit2` I see we don't use releases, right?

joshaber:
>What's the usual update process for `libgit2` I see we don't use releases, right?

No, it's been really ad hoc :disappointed: I'd love to have a more principled process but I doubt that'll happen until libgit2 itself does.

joshaber:
>Does libgit2 no longer support the git protocol?

That would be really surprising but I honestly don't know.

pietbrauer:
@joshaber there was a 0.22.3 release today: https://github.com/libgit2/libgit2/releases/tag/v0.22.3 I guess this does not break APIs. @carlosmn also wrote a comment that 0.23.0 will take a little while still. https://github.com/libgit2/libgit2/issues/3135

joshaber:
Right, sorry, I meant more that since libgit2's still in a high state of flux objective-git will be too.

carlosmn:
Sorry about changing the remotes API so heavily, but the current state was rather unfortunate. There will hopefully also be a similar change to submodules. Back in the day I introduced a pattern very similar to an ORM and that just doesn't really work with Git. These are the changes so we don't have to change anymore (hopefully...).

@phatblat if the url is getting changed, you might be hitting an insteadof rule, we support those now.

phatblat:
Well, @carlosmn was spot on. That last failing test was indeed an `insteadOf` rule I had forgotten about in my `.gitconfig` to get around corporate firewall issues:
screen shot 2015-06-23 at 12 41 26 pm

With that, I believe this PR is ready for review.

joshaber:
That @carlosmn knows his stuff.

Thanks a ton @phatblat!

phatblat:
:tophat:

PR #468, opened 2015-06-25
Adjust GTOID nullability

joshaber:
I don’t see any way that `SHA` could be `nil` without having bigger problems like `malloc` failing. So I think we’re safe to claim it’s not nullable.

/cc @phatblat


mdiep:
:+1: These seem good and right to me.

phatblat:
Cool beans. I was a bit cautious in applying those `nullable` attributes.

joshaber:
>I was a bit cautious in applying those nullable attributes.

:+1: I think that was the right route. We can tighten them up as needed.

PR #469, opened 2015-06-25
Update modulemap

### LINKED ISSUE #467: Update modulemap
I'm seeing a lot of warning with Xcode 6.3.2 like:

```
<module-includes>:1:1: Umbrella header for module 'ObjectiveGit' does not include header '/git2/cred_helpers.h'
```

joshaber:
Mostly fixes #467.

We're still getting:

```
Umbrella header for module 'ObjectiveGit' does not include header '/git2/inttypes.h'
Umbrella header for module 'ObjectiveGit' does not include header '/git2/stdint.h'
Umbrella header for module 'ObjectiveGit' does not include header '/git2/sys/git2/inttypes.h'
Umbrella header for module 'ObjectiveGit' does not include header '/git2/sys/git2/stdint.h'
```

But those are Windows-specific so we don't actually want them. I'm not sure how to silence the warning.


mdiep:
I don't think you're supposed to include any of the `/git2/sys` headers.

joshaber:
Any idea how to silence the warnings about them?

joshaber:
Looks like we should be able to exclude the headers, maybe.

mdiep:
Oh! Xcode is complaining about those because they're being copied into the framework. ObjectiveGit is set up with a folder reference with all the libgit2 headers that's copied into the framework.

The solution is to use a group instead of a reference so you can exclude the headers that we don't want from `ObjectiveGitFramework.xcodeproj` and from the Copy Headers build phase.

mdiep:
(SwiftGit2 does it this way. You should be able to use it as a reference if you need one.)

pietbrauer:
Wouldn't this break integrations for some people that use ObjetiveGit and libgit2 directly?

joshaber:
Interestingly we were already copying the headers. I'm not sure why we had the additional reference.

>Wouldn't this break integrations for some people that use ObjetiveGit and libgit2 directly?

libgit2's headers are still exposed.

This is ready for review again :pray:

mdiep:
> > Wouldn't this break integrations for some people that use ObjetiveGit and libgit2 directly?
>
> libgit2's headers are still exposed.

Some of libgit2's headers are platform-dependent. We're just excluding the ones that don't apply to OS X/iOS.

mdiep:
Xcode is giving me those warnings—even on `master`—but this looks good.

PR #470, opened 2015-06-26
Better unique commit enumeration

joshaber:
This is going to be much faster, though less general, than the existing `-[GTBranch uniqueCommitsRelativeToBranch:error:]` method.

### DIFF
--- a/ObjectiveGit/GTRepository.h
+++ b/ObjectiveGit/GTRepository.h
@@ -565,6 +565,20 @@ NS_ASSUME_NONNULL_BEGIN
 /// Returns whether `ahead` and `behind` were successfully calculated.
 - (BOOL)calculateAhead:(size_t *)ahead behind:(size_t *)behind ofOID:(GTOID *)headOID relativeToOID:(GTOID *)baseOID error:(NSError **)error;
 
+/// Creates an enumerator for walking the local commits in relative to the
+/// tracking branch.
+///
+/// Note that this will only be accurate iff `trackingBranch` truly is
+/// `localBranch`s tracking branch. If you need more general functionality, see
+/// -[GTBranch uniqueCommitsRelativeToBranch:error:].
+///
+/// localBranch    - The local branch.
+/// trackingBranch - `localBranch`s tracking branch.
+/// error          - The error if one occurred.
+///
+/// Returns the enumerator or nil if an error occurred.
+- (nullable GTEnumerator *)enumerateLocalCommitsInBranch:(GTBranch *)localBranch trackingBranch:(GTBranch *)trackingBranch error:(NSError **)error;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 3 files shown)


ObjectiveGit/GTRepository.h
```diff
@@ -565,6 +565,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// Note that this will only be accurate iff `trackingBranch` truly is
+/// `localBranch`s tracking branch. If you need more general functionality, see
+/// -[GTBranch uniqueCommitsRelativeToBranch:error:].
+///
+/// localBranch    - The local branch.
+/// trackingBranch - `localBranch`s tracking branch.
+/// error          - The error if one occurred.
+///
+/// Returns the enumerator or nil if an error occurred.
+- (nullable GTEnumerator *)enumerateLocalCommitsInBranch:(GTBranch *)localBranch trackingBranch:(GTBranch *)trackingBranch error:(NSError **)error;
```
mdiep: `enumerate` seems weird to me since it's returning an enumerator instead of doing the actual work of enumerating. Maybe rename it to `enumeratorForX`?

joshaber: We're using this terminology elsewhere, e.g., `enumerateUniqueCommitsUpToOID`. I'm not opposed to using something else here, but it's also not without precedent.

ObjectiveGit/GTRepository.h
```diff
@@ -565,6 +565,20 @@ NS_ASSUME_NONNULL_BEGIN
 /// Returns whether `ahead` and `behind` were successfully calculated.
 - (BOOL)calculateAhead:(size_t *)ahead behind:(size_t *)behind ofOID:(GTOID *)headOID relativeToOID:(GTOID *)baseOID error:(NSError **)error;
 
+/// Creates an enumerator for walking the local commits in relative to the
+/// tracking branch.
+///
+/// Note that this will only be accurate iff `trackingBranch` truly is
+/// `localBranch`s tracking branch. If you need more general functionality, see
+/// -[GTBranch uniqueCommitsRelativeToBranch:error:].
```
mdiep: I think is only true because of the name you used. AFAICT, it's the same thing as `-[GTRepository enumerateUniqueCommitsUpToOID:relativeToOID:error:]` except that it doesn't do a time sort.

ObjectiveGit/GTRepository.h
mdiep: Yeah, I see that now. :cry:

mdiep:
I think we should give this a more general name, :hocho: `enumerateUniqueCommitsUpToOID:relativeToOID:error:`, and switch `-[GTBranch uniqueCommitsRelativeToBranch:error:]` to this.

ObjectiveGit/GTRepository.h
```diff
@@ -565,6 +555,16 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/// Creates an enumerator for walking the local commits beginning at one OID and
+/// ending at another. It will *not* include the commit for `endingOID`.
+///
+/// startingOID - The starting OID.
+/// endingOID   - The ending OID.
+/// error       - The error if one occurred.
+///
+/// Returns the enumerator or nil if an error occurred.
+- (nullable GTEnumerator *)enumeratorForCommitsStartingAtOID:(GTOID *)startingOID endingAtOID:(GTOID *)endingOID error:(NSError **)error;
```
mdiep: I still don't love this name—I think `endingAt` in particular is a little misleading. :grimacing:

How about `enumeratorForUniqueCommitsFromOID:relativeToOID:` or `enumeratorForCommitsFromOID:excludingCommitsFromOID:`?

mdiep:
:tada:

PR #471, opened 2015-06-27
Cleanup .gitmodules

phatblat:
Tiny change, but the dup entries crash some GUI clients

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,12 +7,12 @@
 [submodule "libssh2"]
 	path = External/libssh2
 	url = https://github.com/libssh2/libssh2.git
-[submodule "Carthage.checkout/Nimble"]
+[submodule "Carthage/checkout/Nimble"]
 	path = Carthage/Checkouts/Nimble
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -7,16 +7,12 @@
...
-[submodule "Carthage.checkout/Nimble"]
+[submodule "Carthage/checkout/Nimble"]
 	path = Carthage/Checkouts/Nimble
 	url = https://github.com/Quick/Nimble.git
-[submodule "Carthage.checkout/Quick"]
+[submodule "Carthage/checkout/Quick"]
 	path = Carthage/Checkouts/Quick
 	url = https://github.com/Quick/Quick.git
-[submodule "Carthage.checkout/xcconfigs"]
+[submodule "Carthage/checkout/xcconfigs"]
```
mdiep: I think these should all be `Carthage/Checkouts/`.

phatblat: :tired_face: thx

mdiep:
Looks great—thanks!

PR #476, opened 2015-07-04
Added overrides for designated initializers in all classes as required by XCode 7 compiler

javiertoledo:
I don’t know if this is the kind of solution you expect, I see it more as a quick patch than a good solution, as there are some warnings about deferred NULL pointers. Any ideas for better defaults for C structs would be really appreciated.

### DIFF
--- a/ObjectiveGit/GTBlame.m
+++ b/ObjectiveGit/GTBlame.m
@@ -21,6 +21,13 @@
 
 @implementation GTBlame
 
+// XCode 7+ require overriding super designated initializers
+- (instancetype)init
+{
+	git_blame *gitBlame = NULL;
+	return [self initWithGitBlame:gitBlame];
+}
+
 - (instancetype)initWithGitBlame:(git_blame *)blame {
 	NSParameterAssert(blame != NULL);
 	
--- a/ObjectiveGit/GTBlameHunk.m
+++ b/ObjectiveGit/GTBlameHunk.m
@@ -12,6 +12,13 @@
 
 @implementation GTBlameHunk
 
+// XCode 7+ require overriding super designated initializers
+- (instancetype)init
+{
+	const git_blame_hunk *hunk = NULL;
+	return [self initWithGitBlameHunk:*hunk];
+}
+
 - (instancetype)initWithGitBlameHunk:(git_blame_hunk)hunk {
 	self = [super init];
 	if (self == nil) return nil;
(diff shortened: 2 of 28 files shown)


phatblat:
Most values passed to the designated initializers should not be null, unless they are documented as an optional value. Object parameters are also tagged with `nonnull` if they are optional, but struct parameters don't have an equivalent attribute. The GT* objects are not going to behave correctly (and will most likely crash) without the required dependencies.

In short, the `-init` method should not be used on most of the objects in this framework. Xcode 7 is highlighting a gap in the API where these methods are still available since they are inherited from `NSObject`.

I had [started on](https://github.com/phatblat/objective-git/commit/b97ad797ef2ff1537f3b8b298a36680aebb8c7f0) some work to make these `-init` methods non-functional, but then came across `NS_ENFORCE_NSOBJECT_DESIGNATED_INITIALIZER` which may actually be the simplest way to deal with this.

joshaber:
Thanks! As a temporary solution goes, this one has a pretty big footprint :anguished: 

I can't find much info on `NS_ENFORCE_NSOBJECT_DESIGNATED_INITIALIZER` besides the fact that it exists. How does it behave?

phatblat:
I don't think there's any docs on `NS_ENFORCE_NSOBJECT_DESIGNATED_INITIALIZER` yet. I found it mentioned in the Dev Forums, but you can see it in the `NSObject.h` file of the OS X 10.11 or iOS 9 SDKs:

screen shot 2015-07-04 at 2 47 26 pm

Looks like it is simply a flag you can use to disable the `NS_DESIGNATED_INITIALIZER` annotation on `-init`. So, it will make Xcode 7 stop complaining about not implementing `-init`, but it won't prevent calls to it.

javiertoledo:
I think that the ideal solution would be to make `init` methods from the superclass unavailable, but it looks like that’s not possible... 

Searching for more info I found this answer from someone at Apple to the same question https://forums.developer.apple.com/message/15575#15575

As we can’t just wipe out the superclass initializers, marking them as “unavailable” and throwing an exception when they’re accessed looks for me like a better approach than just disabling the warnings or implementing methods with `nil` objects that doesn’t make sense...

What do you think? If you like this approach I can close this PR and submit a second one later with the change.

javiertoledo:
I’ve submitted the change I told you about before in PR #477, so I close this one, cheers!

PR #477, opened 2015-07-05
Added NS_UNAVAILABLE flag to all unneeded super initializers (NSObject init)

javiertoledo:
I think that this fix is better than the previously submitted in #476, as XCode marks `init` usages as “unavailable” and, if for some reason an `init` call is achieved in runtime, it raises a runtime error via NSAssert.


mdiep:
This looks like the right fix to me.

phatblat:
:+1:

joshaber:
This is great, thanks so much!

PR #478, opened 2015-07-06
Add failing test for SSH clone

pietbrauer:
Cloning with an ssh URL results in error `Unsupported URL protocol` which is the result of a [`libgit2` check if SSH is enabled](https://github.com/libgit2/libgit2/blob/c7f94123569e8fe00ffb3a35e6a12b6ebe9320ec/src/transport.c#L36-L38). It seems that SSH is not enabled for `objective-git`, [although the `GIT_SSH` flag is set](https://github.com/libgit2/objective-git/blob/master/ObjectiveGitFramework.xcodeproj/project.pbxproj#L1521).

Is this the expected behaviour?

### DIFF
--- a/ObjectiveGitTests/GTRepositorySpec.m
+++ b/ObjectiveGitTests/GTRepositorySpec.m
@@ -84,6 +84,29 @@ describe(@"+cloneFromURL:toWorkingDirectory:options:error:transferProgressBlock:
 		workdirURL = [self.tempDirectoryFileURL URLByAppendingPathComponent:@"temp-repo"];
 	});
 
+	describe(@"with remote ssh repositories", ^{
+		beforeEach(^{
+			originURL = [NSURL URLWithString:@"ssh://github.com/libgit2/rugged.git"];
+			workdirURL = [self.tempDirectoryFileURL URLByAppendingPathComponent:@"rugged"];
+		});
+
+		it(@"should not throw an error", ^{
+			NSError *error = nil;
+			__block BOOL credentialProviderCalled = NO;
+			GTCredentialProvider *provider = [GTCredentialProvider providerWithBlock:^GTCredential * __nonnull(GTCredentialType type, NSString *URL, NSString *username) {
+				credentialProviderCalled = YES;
+				return nil;
+			}];
+
+			repository = [GTRepository cloneFromURL:originURL toWorkingDirectory:workdirURL options:@{ GTRepositoryCloneOptionsCredentialProvider: provider } error:&error transferProgressBlock:transferProgressBlock checkoutProgressBlock:checkoutProgressBlock];
+			expect(repository).notTo(beNil());
+			expect(error).to(beNil());
+			expect(@(transferProgressCalled)).to(beTruthy());
+			expect(@(checkoutProgressCalled)).to(beTruthy());
+			expect(@(credentialProviderCalled)).to(beTruthy());
+		});
+	});
+
 	describe(@"with local repositories", ^{
 		beforeEach(^{
 			originURL = self.bareFixtureRepository.gitDirectoryURL;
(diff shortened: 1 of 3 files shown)


phatblat:
SSH used to work, so this is a regression :frowning:

While I like the idea of having a test like this, there are several reasons it could fail which have nothing to do with the framework. Also, the build server may not even be set up with an SSH key. Perhaps this test could be disabled by default and then manually enabled - or perhaps the logic depends on an environment variable to be present.

I think a better permanent test would be to assert that [git_libgit2_features](https://libgit2.github.com/libgit2/#HEAD/group/libgit2/git_libgit2_features) returns with the `GIT_FEATURE_SSH` bit set.

pietbrauer:
:+1: Will rewrite it then and find a test file where it suites in.

phatblat:
You can [disable](https://github.com/Quick/Quick/blob/swift-2.0/Documentation/QuickExamplesAndGroups.md#temporarily-disabling-examples-or-groups) the test with `xdescribe` or `xit`. That test will be a great help in verifying a fix for the SSH functionality (which I suspect is build related).

pietbrauer:
Yeah, it seems that `GIT_SSH` is not set when [`script/update_libgit2`](https://github.com/libgit2/objective-git/blob/master/script/update_libgit2#L25) is being built. Too I am lacking a ton of `cmake` knowledge.

phatblat:
I've worked with the build scripts and cmake a bit so I can look into this issue.

pietbrauer:
I would also love to work on this but have little hope. I tried adding `-DCMAKE_C_FLAGS="${CMAKE_C_FLAGS} -DGIT_SSH"` but that did not help.

phatblat:
The libgit2 cmake build will disable SSH support if it doesn't find the libraries, which may be what's going on here. Are you seeing this when building for Mac, iOS or both?

pietbrauer:
Both as Mac is the test target and my iOS app also fails. Afaik the Mac build uses system OpenSSH and the iOS app depends on the OpenSSH iOS build.

pietbrauer:
@phatblat I fixed the build and enabled SSH support. I also refactored the test to test for the correct error (SSH credentials not there).

phatblat:
Fantastic! :balloon: I'll test this out later this evening. I have an idea for getting those SSH credentials working.

pietbrauer:
Superb!! 🎉

pietbrauer:
@phatblat Did you have any chance to get to the tests or maybe share your approach?

phatblat:
Sorry, got tied up the last couple evenings.

I started adding some logic to pick up the SSH keys from `~/.ssh`, but then I noticed that there is already an [SSH clone test](https://github.com/libgit2/objective-git/blob/master/ObjectiveGitTests/GTRepositorySpec.m#L139-L178). This test doesn't run unless you have the `GTUserName`, `GTPublicKey`, `GTPrivateKey` and `GTPrivateKeyPassword` defined in the current environment.

screen shot 2015-07-09 at 6 56 33 pm

I presume these values are defined on the GitHub build server. I defined them as above and ran this test on my Mac. So, SSH is working fine on the Mac side - without these changes.

phatblat:
It's been challenging validating this change on iOS as my client app builds with Xcode 7 but I'm not able to build ObjectiveGit from source with 7. However, I built this branch with 6.4 and copied over the .framework and the `GIT_FEATURE_SSH` flag is coming back set from `git_libgit2_features`.

This looks like progress, but I'd like to do some deeper validation because this change forces `GIT_SSH` to be set, but that flag is supposed to automatically get set by cmake if it finds the SSH lib. I'm concerned about the lib not being found since that can cause issues when an app tries to actually use SSH.

phatblat:
I think this may be the real fix: phatblat@aed2cc39c831697102ac6aff317990729b55dd2c. I misspelled `LIBSSH2_INCLUDE_DIRS` when I added that to the libgit2 build scripts.

pietbrauer:
I can verify it, my app still runs on Xcode 6.4. Will try it out today. If it fixes the issue, I will cherry pick and rebase this branch.

phatblat:
Also check out #483 to see if that works for you.

pietbrauer:
Closing in favour of #483

PR #479, opened 2015-07-06
Give extern'd string known nullability.

joshaber:
Moved `NS_ASSUME_NONNULL_BEGIN` to before the definition of our `extern`’d strings.

Otherwise framework consumers would see warnings about unspecified nullability.

### DIFF
--- a/ObjectiveGit/GTDiff.h
+++ b/ObjectiveGit/GTDiff.h
@@ -14,6 +14,8 @@
 @class GTRepository;
 @class GTTree;
 
+NS_ASSUME_NONNULL_BEGIN
+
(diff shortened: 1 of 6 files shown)


phatblat:
Thank you! I've been meaning to do this.

joshaber:
@phatblat Given that you're a contributor, feel free to assign yourself and merge :smile:

phatblat:
On it. Now, if only I can type `nullabilly` right :wink:

joshaber:
:trollface:

phatblat:
Looks great! :palm_tree:

PR #487, opened 2015-07-11
Ignore Carthage/Build

phatblat:
Running `carthage update` builds by default which dirties the work tree. While the current project configuration doesn't require these dependencies to be built with Carthage (Quick and Nimble are just added to the workspace as dependent projects), I always forget to add `--no-build` and am left wondering why the work tree is dirty.


joshaber:
:+1:

PR #489, opened 2015-07-11
Don't email about build status.

joshaber:
Build status is reported on the PR so no need to annoyingly email.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ osx_image: xcode6.4
 language: objective-c
 install: script/bootstrap
 script: script/cibuild
+notifications:
+  email: false


pietbrauer:
Yeah, wasn't sure because I removed the campfire notifications as well. I quite like the mails because the build takes about 20 minutes so you might do something else in the mean time and forget to check back. They also don't email, when the build passes afaik.

joshaber:
Yeah, I get that but I think we should opt for not bothering contributors. I wish people could opt-in but it doesn't look like TravisCI supports anything like that :(

phatblat:
Travis does have an [on_success](http://docs.travis-ci.com/user/notifications/#Email-notifications) option.

phatblat:
I agree about not bothering contributors. @pietbrauer if you want all notifications you could add your email to the config.

pietbrauer:
Fine with me.

phatblat:
BTW, I love seeing the CI status here in the PR, at least the Travis ones that I can drill down on.

We might be able to use Slack to bounce build notifications through. Then interested parties could opt into build notifications using Slack.

PR #464, opened 2015-06-17
Pull

phatblat:
Initial Pull implementation preferring fast-forward merge if possible or a "normal" merge if branches have diverged, depending on the merge analysis. Pull will fail if a conflict is detected.

### TODO

* [x] Fast-forward merge
* [x] Normal merge
* [x] Merge conflict communication to caller (via error)
* [x] Tests

### Out of scope
 
* Pull with rebase
* Explicit merge commit (aka `--no-ff`)
* Prevent merge commit (aka `--ff-only`)
* Pull an unborn branch

PR #494, opened 2015-07-21
Enable verbose build output from iOS build scripts

phatblat:
An attempt to work around Travis build timeouts due to lack of output from the build (mentioned in #464).

Let's see if Travis succeeds on this PR 🙏🏽

### DIFF
--- a/ObjectiveGitFramework.xcodeproj/project.pbxproj
+++ b/ObjectiveGitFramework.xcodeproj/project.pbxproj
@@ -1205,7 +1205,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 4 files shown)


joshaber:
It would seem not :cry:

pietbrauer:
My suggestion would be to run `script/update_lib*` manually before we start `xctool` this way it produces output & runs fast with `xctool` (because the libraries already exist). I could take care of it unless @phatblat really really wants to do it. I don't think the verbose mode is needed.

phatblat:
@pietbrauer got for it. There's an issue with this change where Xcode treats some output from configure as an error.

phatblat:
I'm going to close this until I can figure out a good way to deal with that configure output. Don't want to hide it but don't want Xcode to think it's an error.

PR #495, opened 2015-07-22
Prebuild OpenSSL in order to workaround Travis timouts

pietbrauer:

PR #496, opened 2015-07-22
Bump libgit2

joshaber:
Pick up the fix for https://github.com/libgit2/libgit2/issues/3318 so that our tests can run.


joshaber:
(It's not at all clear why this failed, but presumably the same timeout problem #495 should help address.)

PR #497, opened 2015-07-29
Just install any missing tools in bootstrap

phatblat:
It seems silly for the bootstrap script to fail if xctool and cmake are not installed but then to go ahead and install other things through Homebrew.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -104,7 +104,9 @@ bootstrap_submodule ()
 
 update_submodules ()
 {
-    git submodule sync --quiet && git submodule update --init && git submodule foreach --quiet bootstrap_submodule
+    git submodule sync --quiet && \
+        git submodule update --init && \
+        git submodule foreach --quiet bootstrap_submodule
 }
 
 export -f bootstrap_submodule
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -9,7 +9,8 @@ export SCRIPT_DIR=$(dirname "$0")
 config ()
 {
     # A whitespace-separated list of executables that must be present and locatable.
-    : ${REQUIRED_TOOLS="xctool cmake"}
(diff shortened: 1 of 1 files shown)


mdiep:
I'd rather not increase our dependency on homebrew. Could we just print out a command to install through homebrew instead, to let the user have the option? I think it's kinda rude to install things like this anyway.

phatblat:
Running `brew install xctool cmake libssh2 libtool autoconf automake pkg-config` would do it.

phatblat:
So, the bootstrap script currently does 4 things:
1. Stop with error if xctool or cmake are not installed (but they are available through Homebrew)
2. Install libssh2 libtool autoconf automake pkg-config through Homebrew
3. Symlink libssh2 library and headers into /usr/local if `brew --prefix` is something nonstandard
4. Update submodules

Right now it's heavily dependent on Homebrew.

mdiep:
I don't have strong feelings. I don't use homebrew, so it probably doesn't affect me either way. :smile:

phatblat:
Made a couple minor changes after taking @mdiep's POV into account
* Abort script early with helpful message if Homebrew isn't installed
* Update submodules first in case script aborts

Here's what the message looks like if Homebrew isn't installed:

```
$ script/bootstrap
*** Updating submodules...
*** Checking dependencies...

Homebrew is not installed (http://brew.sh). You will need to manually ensure the following tools are installed:
  xctool cmake libssh2 libtool autoconf automake pkg-config

Additionally, the following libssh2 files must be symlinked under /user/local:
  lib/libssh2.a include/libssh2.h include/libssh2_sftp.h include/libssh2_publickey.h
```

mdiep:
This looks great to me, but someone else who has Homebrew installed should test it out. :+1:

joshaber:
:metal:

PR #500, opened 2015-08-07
Declared some extra nullability to please Xcode 7 beta.

wilshipley:
I added some "nonnull" and "_Null_unspecified" attributes to four public headers.

This shouldn't change the functionality at all, but it removes some warnings I get when I compile on the latest Xcode.

This is my first pull request on this project, apologies in advance if I'm doing this totally wrong.

### DIFF
--- a/ObjectiveGit/GTConfiguration+Private.h
+++ b/ObjectiveGit/GTConfiguration+Private.h
@@ -18,6 +18,6 @@
 /// repository - The repository in which the config resides. May be nil.
 ///
 /// Returns the initialized object.
-- (nullable instancetype)initWithGitConfig:(git_config *)config repository:(nullable GTRepository *)repository NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithGitConfig:(nonnull git_config *)config repository:(nullable GTRepository *)repository NS_DESIGNATED_INITIALIZER;
 
 @end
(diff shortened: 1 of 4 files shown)


phatblat:
__null_unspecified

wilshipley:
Ok, done! Hopefully they don't deprecate "__null_unspecified" any time soon. (I don't think they're eager to.)

phatblat:
Fantastic! Thanks for the contribution. 

If they end up dropping the nullability attributes introduced in Xcode 6.3 we will probably need our own compatibility macro. But I'd prefer not to get into that unless it ends up being recessary.

wilshipley:
Yah, no need to go looking for trouble.

I have another thing I did where I made an Xcode project for this and for libgit2 that you can include as a subproject in another Xcode project, and it'll build on iOS or OS X automatically all inside Xcode with no external tools (e.g. no cmake, no scripts). Not sure if you guys want that, though, because the libgit2 one DOES require maintaining by the Xcode project by hand if you add / remove files.

phatblat:
I'm very interested in seeing this Xcode project got libgit2. Do you have it in a branch on your fork of ObjectiveGit?

I understand at least some of the issues with having a build system separate from the libgit2 cmake build, but there are a couple issues I've been looking into that I can't seems to resolve:
* git2 is not a real module, it's a submodule under ObjectiveGit 
* bitcode support - this is pretty easy to add for one architecture, but gets dropped when we lipo the slices together

Both of these seem easy to solve with a real Xcode build.

joshaber:
:metal: Those warnings had been bugging me.

PR #459, opened 2015-04-10
Wrap all the various way of doing checkouts in GTCheckoutOptions

tiennou:
Fixes #457

PR #501, opened 2015-08-09
Add stash apply & pop support

tiennou:
Sorry, there's no test in there because I wasn't running the greatest and latest Xcode when I wrote that, and time is pretty scarce yadda yadda.

### DIFF
--- a/ObjectiveGit/GTRepository+Stashing.h
+++ b/ObjectiveGit/GTRepository+Stashing.h
@@ -19,6 +19,14 @@ typedef NS_OPTIONS(NSInteger, GTRepositoryStashFlag) {
 	GTRepositoryStashFlagIncludeIgnored = GIT_STASH_INCLUDE_IGNORED
 };
 
+/// Flags for -applyStashAtIndex:flags:error: and
+/// -popStashAtIndex:flags:error.
+/// Those can be ORed together. See git_apply_flags for additional information.
+typedef NS_OPTIONS(NSInteger, GTRepositoryApplyFlag) {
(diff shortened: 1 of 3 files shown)


joshaber:
Those build failures look legit. Looks like the PR was written against a different version of libgit2.

tiennou:
Oh my fault, I failed to build the correct target before pushing, and the stash API now has a "nice" `git_stash_options` struct. I'll update the API and try to write some of those tests then...

joshaber:
Sorry, I totally missed that this was ready for re-review.

It looks good to me. :+1: Just wanna confirm that this is done from your point of view too.

tiennou:
My only grudge against it is that it's untested (like, I didn't actually use it, just wrote the wrappers). I *might* add test cases soon-ish, but if you want to merge, please do.

joshaber:
Oh lol I totally didn't notice that. Tests would be good :+1:

tiennou:
Here are some tests. I also made a little tweak to the signatures because I noticed those pesky error pointers weren't at the "right" place...

Note that this is not designed to handle conflicts at the moment, since `git_stash_options` embeds a `git_checkout_options` structure, this would make a messy parameter list. Related to #459. So the path forward here is either that other PR, or providing
```
- applyStashAtIndex:flags:checkoutStrategy:checkoutNotify:error:stashProgressBlock:checkoutProgressBlock:checkoutNotifyBlock:
```

:worried:

joshaber:
Ugh yeah, I think that PR is our best way forward on that front.

This looks good :metal:

PR #504, opened 2015-08-09
Give the dependency target run scripts real names

phatblat:
This is a tiny little change, but it helps to know "what's taking so long" when building for iOS, especially when you're following a log from Carthage or Travis.

screen_shot_2015-08-09_at_10_15_02_am

The run scripts that simply call a script have been renamed to match the target name.

screen_shot_2015-08-09_at_10_16_45_am


mdiep:
:+1:

PR #509, opened 2015-08-18
Fix checkout methods utilizing notifyBlock/GTCheckoutStrategySafe

dleehr:
Currently, the `performCheckout` methods move __HEAD__ to the new ref/commit before calling `git_checkout_head`. This is incorrect behavior.

When `git_checkout_head` is called, it reports changes between the __workdir__ and __HEAD__ per-file to the `notifyBlock`, allowing the block to abort the checkout based on a reason (dirty/conflict/etc).

However, since __HEAD__ has already been moved to a different ref/commit, the `notifyBlock` is notified about changes between the current __workdir__ and the requested ref/commit. This results in a dirty state for any file that differs between the old ref and the new ref, regardless of whether or not any change was made in the workdir.

The correct behavior (e.g. with `GTCheckoutStrategySafe`) is to notify of changes that will cause data loss in the workdir.

This change leaves __HEAD__ in position until after the checkout, correcting the checkout behavior. When `GTCheckoutStrategySafe` is used, `notifyBlock` is only invoked as expected (e.g. dirty files that would be overwritten on checkout). Also, __HEAD__ is only moved if the checkout succeeds, preventing other confusing behavior.


dleehr:
Actually this is not as straightforward as I thought. `performCheckoutWithStrategy` calls `git_checkout_head()`, which is why it expects HEAD to be moved first.

I think the correct behavior is to determine the target __tree__ and use `git_checkout_tree()`

PR #510, opened 2015-08-19
Fix checkout methods that use notifyBlock/GTCheckoutStrategySafe

dleehr:
Previously, `performCheckout` method moved __HEAD__ to the target, then called `git_checkout_head()`. This does not match behavior in git core, and prevents the `checkoutStrategy` and `notifyCallback` parameters from being used correctly. The only way to reliably checkout a branch or commit was to use GTCheckoutStrategyForce, which throws away conflicts. Also, the HEAD could be moved even if the checkout fails, resulting in a confused state.

This PR corrects the checkout behavior with two changes:

1. __HEAD__ is not moved until after the checkout succeeds.
2. `git_checkout_tree()` is used instead of `git_checkout_head()`. The tree that is checked out is the target object (commit or ref).

Included tests demonstrate the correct behavior catching a conflicted file. The notifyBlock is called to report the conflict, and with GTCheckoutStrategySafe, the checkout fails because of the conflict as expected.

### DIFF
--- a/ObjectiveGit/GTRepository.m
+++ b/ObjectiveGit/GTRepository.m
@@ -809,7 +809,7 @@ static int checkoutNotifyCallback(git_checkout_notify_t why, const char *path, c
 	return gitError == GIT_OK;
 }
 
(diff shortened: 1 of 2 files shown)


phatblat:
Nice work! :space_invader:

PR #512, opened 2015-08-24
Remove deprecated targetCommitAndReturnError

phatblat:
Conflicts with the non-deprecated `targetCommitWithError:` method in Swift 2 (Xcode 7b6)

```
error: ambiguous use of 'targetCommit()'
                    let gt_commit = try! gt_branch.targetCommit()
                                         ^
ObjectiveGit.GTBranch:38:14: note: found this candidate
  @objc func targetCommit() throws -> GTCommit
             ^
ObjectiveGit.GTBranch:86:14: note: found this candidate
  @objc func targetCommit() throws -> GTCommit
```


phatblat:
If we want to keep `targetCommitAndReturnError`, another solution is to mark it as [NS_SWIFT_UNAVAILABLE](https://gist.github.com/fastred/d92ca6145cc2b0cbf6f0) so that it is not visible from Swift.

joshaber:
:fire:

PR #513, opened 2015-08-25
Made two -moveHEAD... methods public.

wilshipley:
Moved private method declarations -moveHEADToReference:error: and -moveHEADToCommit:error: to the header so they can be used from apps. This seems important because I'm told the only legitimate way to move the HEAD is with these methods (and the underlying git2 functions) and NOT to delete and recreate the HEAD reference manually, so these methods must be public if client apps need to move the HEAD.


External/libgit2
```diff
@@ -1 +1 @@
-Subproject commit c27b4afcdd80f5a45d7120044bf3d78272181abb
+Subproject commit abbe17ef1ffbc264101808829a3317c1c805df96
```
mdiep: It looks like you've accidentally moved libgit2 to an earlier revision.

wilshipley:
Ugh I suck.

wilshipley:
Ok, I definitely don't understand how to add submodules to submodules in git, which is ironic, since I'm trying to add code to git itself.

phatblat:
Submodules are hard. In your branch you want to drill into the `External/libgit2` folder, then check out c27b4af. Pop back up to objective-git and then check in the change to `External/libgit2` (the new value should be c27b4af)

PR #514, opened 2015-08-26
Fix bug in -isWorkingDirectoryClean

dleehr:
Comment described checking for untracked files, but status was compared to GTStatusDeltaStatusAdded instead of GTStatusDeltaStatusUntracked


phatblat:
Looks good to me.

PR #515, opened 2015-08-26
2nd attempt: Hopefully just requesting pull of GTRepository.h

wilshipley:
Moved private method declarations -moveHEADToReference:error: and -moveHEADToCommit:error: to the header so they can be used from apps. This seems important because I'm told the only legitimate way to move the HEAD is with these methods (and the underlying git2 functions) and NOT to delete and recreate the HEAD reference manually, so these methods must be public if client apps need to move the HEAD.

### DIFF
--- a/ObjectiveGit/GTRepository.h
+++ b/ObjectiveGit/GTRepository.h
@@ -269,6 +269,14 @@ extern NSString * const GTRepositoryInitOptionsOriginURLString;
 /// Returns a GTReference or nil if an error occurs.
 - (nullable GTReference *)headReferenceWithError:(NSError **)error;
 
+/// Move HEAD reference safely, since deleting and recreating HEAD is always wrong
+///
+/// error - If not NULL, set to any error that occurs.
+///
+/// Returns NO if an error occurs.
+- (BOOL)moveHEADToReference:(GTReference *)reference error:(NSError **)error;
+- (BOOL)moveHEADToCommit:(GTCommit *)commit error:(NSError **)error;
+
 /// Get the local branches.
 ///
 /// error - If not NULL, set to any error that occurs.


External/libgit2
```diff
@@ -1 +1 @@
-Subproject commit c27b4afcdd80f5a45d7120044bf3d78272181abb
+Subproject commit abbe17ef1ffbc264101808829a3317c1c805df96
```
mdiep: This is still reverting libgit2 to an older version. You should be able to follow the instructions [here](https://github.com/libgit2/objective-git/pull/513#issuecomment-134785288) to push a new commit to this PR.

ObjectiveGit/GTRepository.h
```diff
@@ -269,6 +269,14 @@ extern NSString * const GTRepositoryInitOptionsOriginURLString;
 /// Returns a GTReference or nil if an error occurs.
 - (nullable GTReference *)headReferenceWithError:(NSError **)error;
 
+/// Move HEAD reference safely, since deleting and recreating HEAD is always wrong
+///
+/// error - If not NULL, set to any error that occurs.
+///
+/// Returns NO if an error occurs.
+- (BOOL)moveHEADToReference:(GTReference *)reference error:(NSError **)error;
+- (BOOL)moveHEADToCommit:(GTCommit *)commit error:(NSError **)error;
```
mdiep: Can you add the documentation to the second method as well? Even if it's the same, Xcode won't show it for the second method unless it's repeated.

mdiep:
A couple small issues, but I think this would be a fine thing to do.

FWIW, there's no need to close the pull request and open a new one. You can just make the changes, push them, and the PR will update. Just be sure to leave a comment so we know to take another look. :smile:

wilshipley:
Ok, I need to stop doing the pull requests from the GitHub Desktop app because it's hard to see the complete diffs in it. I just looked at the GitHub web interface and it's MUCH more straightforward.

I think part of why I'm fighting with this so much is that in my project I have Objective-Git and libgit2 as peer submodules (both are in the same subdirectory of my main project, not one inside the other) inside my main app repository, whereas in this project libgit2 is a submodule of (and a directory or two below) Objective-Git.

Muddling through...

joshaber:
#535 tackled this. Thanks!

PR #517, opened 2015-08-28
Make path in checkoutProgressBlock optional

alehed:
The path from the checkout progress block is set to nil when
called the first time from git_checkout_iterator. This happens
when calling report_progress() on line 2547 in checkout.c. This
in turn causes a crash.


phatblat:
Nice find! :tada:

PR #519, opened 2015-09-14
Fix issue when compiling when UIKit is imported

misterfifths:
I've gotten Objective-Git working with Cocoapods for iOS, but this one small change is necessary. It seems like Cocoapods unconditionally includes UIKit in the prefix header for the target it generates, and doing so confuses this code -- it thinks the `path` selector you're using is one of the ones from CoreAnimation, and expects it to return a `CGPathRef`. The explicit cast fixes that problem.

### DIFF
--- a/ObjectiveGit/GTRepository.m
+++ b/ObjectiveGit/GTRepository.m
@@ -134,7 +134,7 @@ typedef struct {
 	[optionsDict[GTRepositoryInitOptionsMode] unsignedIntegerValue];
 	options.workdir_path = [optionsDict[GTRepositoryInitOptionsWorkingDirectoryPath] UTF8String];
 	options.description = [optionsDict[GTRepositoryInitOptionsDescription] UTF8String];
-	options.template_path = [optionsDict[GTRepositoryInitOptionsTemplateURL] path].UTF8String;
+	options.template_path = ((NSURL *)optionsDict[GTRepositoryInitOptionsTemplateURL]).path.UTF8String;
 	options.initial_head = [optionsDict[GTRepositoryInitOptionsInitialHEAD] UTF8String];
 	options.origin_url = [optionsDict[GTRepositoryInitOptionsOriginURLString] UTF8String];
 


phatblat:
While this change is certainly innocuous, I'm worried that it may get factored out in the future since it looks like an unnecessary cast.

Are you working on a .podspec for ObjectiveGit? If so, I'd like to see that and entertain the idea of officially supporting CocoaPods. If we were to work a `pod lib lint` into the [`cibuild`](https://github.com/libgit2/objective-git/blob/master/script/cibuild) script then it would help catch any regression in CocoaPods support.

misterfifths:
Here's my podspec for 0.8.7: https://gist.github.com/misterfifths/534e62072f703e0947bc. It's rough around the edges -- I took a skeleton from somewhere or other (some googling of "objective-git podspec") and fixed it to actually work. I haven't tested the OSX side of it at all.

phatblat:
Nice, thanks for sharing! For a while I was tinkering with my own [`ObjectiveGit.podspec`](https://github.com/phatblat/objective-git/blob/ben/archive/podspec-0.2.004/ObjectiveGit.podspec) but never got it far enough along to get it to build right. I'll review your spec later this week.

joshaber:
>While this change is certainly innocuous, I'm worried that it may get factored out in the future since it looks like an unnecessary cast.

Indeed, perhaps we should link to this PR in a comment?

tomj:
Hey everyone I'm happy to take a look at this, have been looking for somewhere to get involved with objective-git - might check in with the folks over at CCP to find out about that default UIKit include too

tomj:
For completeness the folks over at CocoaPods [discussed the UIKit import issue a while ago](https://github.com/CocoaPods/CocoaPods/issues/1691) and closed it; you can still see it in the CCP source [here](https://github.com/CocoaPods/CocoaPods/blob/691e3de4405a0b7174ccd9ca85912a045d286aa2/lib/cocoapods/generator/header.rb#L86).  Will dig around and see what I can come up with to avoid this issue without having to add a PR link comment or similar - I'm hoping there's a more elegant solution 😎

pietbrauer:
@misterfifths Is this still an issue with CocoaPods `1.0.beta2`?

misterfifths:
@pietbrauer Just checked, and it is, though I must admit I don't understand why... doesn't seem like the blanket UIKit import changed in the prefix header.

pietbrauer:
@misterfifths So it is still an issue and this pull request is still needed to make your `podspec` work with CocoaPods `1.0.0.beta2`? I am just asking for sanitising reasons.

misterfifths:
No, it seems all is well as of 1.0.0.beta2, so this can probably be closed.

pietbrauer:
Cool, thanks. If not, feel free to reopen.

PR #522, opened 2015-10-02
Adjusted GTMergeAnalysis type

slavikus:
GTMergeAnalysis is actually NS_OPTIONS and not NS_ENUM. Swift is getting confused at the bitmasks when it's defined as NS_ENUM.


joshaber:
Good catch :+1:

PR #516, opened 2015-08-28
[Xcode 7] Add types to collections in GT headers

JrGoodle:
This adds support for generics with typed collections added in Xcode 7. The `.travis.yml` is updated to use the Xcode 7 beta images.

PR #523, opened 2015-10-05
Xcode 7 + silence stupid warning

joshaber:
Silences the warnings like:
```
<module-includes>:1:1: warning: umbrella header for module 'ObjectiveGit' does not include header blah blah blah blah
```


joshaber:
Carthage is really struggling with this for me:

A bunch of these:

```
Caught signal triggered by the Swift runtime!
Illegal instruction: 4
```

and then also:

```
*** Skipped building ZipArchive due to the error:
Dependency "ZipArchive" has no shared framework schemes
```

I'm not sure if these are Carthage bugs ... or ... computer bugs ...

mdiep:
I think you should be using `--use-submodules --no-build`, in which case you should never see that message.

joshaber:
Ohhhh right :+1:

phatblat:
FYI, I've been dabbling on a real fix for those module warnings/errors in phatblat/objective-git#1. The change is to build libgit2 as its own "git2" module instead of the "ObjectiveGit/git2" submodule that we're maintaining now. The result is [git2.framework](https://github.com/phatblat/git2) but I haven't gone as far as including OpenSSL and libssh2 yet.

joshaber:
That'll be :sparkles: if it works!

phatblat:
There are some other changes in #516 to get the build to work on Travis. Specifically, a newer version of xctool is necessary (e95cd59f5).

joshaber:
Indeed :disappointed: 

I opened https://github.com/facebook/xctool/issues/605 to see when they expect a new release.

phatblat:
Cool, thanks. I was starting to think we should just live with xctool at HEAD and open an issue to fix it later. If they can push out a release that will be even better.

JrGoodle:
@joshaber thanks for opening that issue, looks like it's coming "in a week or so" https://github.com/facebook/xctool/issues/605#issuecomment-147765541

I've been keeping watch for the new release; I'll update #516 as soon as I see it's released

joshaber:
This is ready now :pray:

mdiep:
:ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship:

Issue #441, opened 2015-02-15
Add 0.5.1 zip to release


pietbrauer:
In the most recent carthage release it is possible to attach zips. I tried to build it for 0.6.0 but it is not building because of a libssh2 error.

However I was able to build 0.5.1 and tried to attach it to this ticket, but it is 12,3 MB and Github only allows 10 MB. It would be so cool if I could attach it to the release :cat2:

Many Thanks!

jspahrsummers:
I've just invited you to be a collaborator. :blush:

pietbrauer:
I uploaded the binary and since removed it again because the header search path is not working.

So everyone including this project via carthage will get the `git2/filter.h file not found` errors. Because obviously `"$(SRCROOT)/Carthage/Checkouts/objective-git/External/libgit2/include/"` does no longer exist.

Any idea?

jspahrsummers:
Bleh, this is the problem that the [module map](https://github.com/libgit2/objective-git/pull/438) is intended to solve, but that only came in 0.6.

pietbrauer:
Even with the module map it is not solved (tried with 0.6) and the mac build needs to set `HEADER_SEARCH_PATHS` too so it can't be an iOS configuration error. Any ideas?

jspahrsummers:
Sorry, I don't have a whole lot of free time to investigate this, especially because ObjectiveGit is successfully building in our subproject setup. :confused:

pietbrauer:
Sure, maybe @phatblat has an idea. He already had a fix every time I opened an issue :smiley:

jspahrsummers:
Now that #442 is fixed, I was able to build and upload a binary for [0.6.1](https://github.com/libgit2/objective-git/releases/tag/0.6.1). Does that one work?

pietbrauer:
Worked but had to add `$(SRCROOT)/Carthage/Build/iOS/ObjectiveGit.framework/Headers/` to `HEADER_SEARCH_PATHS`

jspahrsummers:
I wonder if we could fix that by using angle-bracket imports in the module map? Anyways, glad to hear it's working now.

pietbrauer:
You mean like [this](https://github.com/pietbrauer/objective-git/commit/0d746ab8e4d4a177464b70cc76a11ceee1ca0bd9)? That doesn't work.

jspahrsummers:
Ah, alright. :disappointed: 

Thanks for trying!

zetachang:
[...]

PR #525, opened 2015-10-15
Added note setup step

aral:
I had to manually add a header search path as per https://github.com/libgit2/objective-git/issues/441 in order to compile. Added that as a step to the instructions. Only tested on iOS so added to end of that list. Not sure if it also applies for Mac targets (assuming so). If so, please feel free to move to the top where it might make more sense and apply to both.


phatblat:
Thanks! I forgot about this step.

aral:
No worries ;) And thanks!

Issue #524, opened 2015-10-08
Should `isAGitDirectory` be public?


samskiter:
I can't find any use of it. Should it be a public class method of `GTRepository`?

joshaber:
Hah, yeah, and it's almost certainly gonna be wrong in a lot of situations. A PR to :fire: it would be great :smile:

samskiter:
Ah I see. It's an old method. it's not currently public, i just wondered if it should be as it's pretty useful

joshaber:
It seems likely that libgit2 has some function like that that we should use instead.

pietbrauer:
What I did in my app is to check if [`[GTRepository initWithURL:error:]`](https://github.com/libgit2/objective-git/blob/master/ObjectiveGit/GTRepository.m#L179) and check for the `libgit2` error which is returned if it is not a repository. I tried both ways, the `isAGitDirectory` check and checking for the error, they both result in the same behaviour.

joshaber:
Yeah, I'd expect `-initWithURL:error:` to be more future-safe.

PR #527, opened 2015-10-21
Remove `isAGitDirectory:`

pietbrauer:
As discussed in https://github.com/libgit2/objective-git/issues/524 it is not public and `[GTRepository initWithURL:error:]` should be more future safe.


joshaber:
:sparkling_heart:

PR #534, opened 2015-11-10
Fix a few compile errors in the iOS Tests target

phatblat:
I guess we're not running these on travis...


pietbrauer:
Looks legit but I am unsure why we aren't running iOS tests on CI.

pietbrauer:
Tests compile and run at least :+1:

PR #535, opened 2015-11-10
Expose -moveHEAD

phatblat:
Expose move HEAD operations through public API. Includes tests.

> Moved private method declarations -moveHEADToReference:error: and -moveHEADToCommit:error: to the header so they can be used from apps. This seems important because I'm told the only legitimate way to move the HEAD is with these methods (and the underlying git2 functions) and NOT to delete and recreate the HEAD reference manually, so these methods must be public if client apps need to move the HEAD.

Supersedes #515 from @wilshipley

### DIFF
--- a/ObjectiveGit/GTRepository.h
+++ b/ObjectiveGit/GTRepository.h
@@ -280,7 +280,7 @@ extern NSString * const GTRepositoryInitOptionsOriginURLString;
 /// Move HEAD reference safely, since deleting and recreating HEAD is always wrong.
 ///
 /// commit - The commit which HEAD should point to.
-/// error - If not NULL, set to any error that occurs.
+/// error  - If not NULL, set to any error that occurs.
 ///
 /// Returns NO if an error occurs.
 - (BOOL)moveHEADToCommit:(GTCommit *)commit error:(NSError **)error;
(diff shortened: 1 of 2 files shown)


ObjectiveGit/GTRepository.h
```diff
@@ -269,6 +269,22 @@ extern NSString * const GTRepositoryInitOptionsOriginURLString;
 /// Returns a GTReference or nil if an error occurs.
 - (nullable GTReference *)headReferenceWithError:(NSError **)error;
 
+/// Move HEAD reference safely, since deleting and recreating HEAD is always wrong.
+///
+/// reference - The new target reference for HEAD.
```
joshaber: Style nit: could you align the `-` for the args?

joshaber:
Just a style nit. Thanks!

phatblat:
you got it dude!

ObjectiveGit/GTRepository.h
joshaber: Looks like this was was missed :smile:

phatblat: :astonished:

phatblat:
so sorry

joshaber:
:metal:

PR #538, opened 2016-01-22
Add list of files to merge conflict error

pietbrauer:
Currently we only return `Merge conflict, pull aborted` when a merge conflict is detected while pulling this enumerates the conflicted files and returns `Merge conflict in files: index.html, _includes/head.html. Pull aborted.`.

Enumeration should be ok as it is most likely run in a background thread anyway. What do you think?

### DIFF
--- a/ObjectiveGit/GTRepository+Pull.m
+++ b/ObjectiveGit/GTRepository+Pull.m
@@ -17,6 +17,7 @@
 #import "GTRepository+RemoteOperations.h"
 #import "GTTree.h"
 #import "NSError+Git.h"
+#import "GTIndexEntry.h"
 #import "git2/errors.h"
 
 @implementation GTRepository (Pull)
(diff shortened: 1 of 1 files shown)


tiennou:
Maybe store that in the error's user info ?

pietbrauer:
Yeah true, will fix that.

joshaber:
I'm inclined to say that we should leave doing this in the users' hand. They may very well want a different error message, want it localized, etc.

tiennou:
Depends. It seems the index where conflicts originated is not accessible externally, so you'll have to perform the pull by hand afterward to get the conflicted list.

pietbrauer:
Yep, exactly My intention. To get the list of files we have to rebuild the pull feature.

joshaber:
Eh, yeah, fair enough.

pietbrauer:
@joshaber I will do another PR that sets the conflicted files in the user info?

joshaber:
Yep that's be great!

PR #539, opened 2016-01-22
+[GTDiff diffOldTree:withNewIndex:…]

slavikus:
Adding support for diffing a tree and an arbitrary GTIndex (for in-mem comparisons of pending merges)


joshaber:
Seems legit. Thanks!

PR #540, opened 2016-01-26
Added missing nullable keyword to [GTIndex entryWithPath:path] convenience method

onecrayon:
Missing keyword made it impossible to use this method properly in Swift (and since it basically just aliases entryWithPath:error: with a NULL error, there's no reason the signature should be different).


joshaber:
Ah, indeed. Thanks!

PR #543, opened 2016-01-27
Fixed erroneous shouldIgnoreFile return value

onecrayon:
Per the [git_status_should_ignore](https://libgit2.github.com/libgit2/#HEAD/group/status/git_status_should_ignore) docs (emphasis mine):

> ignored: Boolean returning 0 if the file is **not ignored**, 1 if it is

However, the current implementation was returning YES ("the file should be ignored") if the ignore state was 0 ("the file should not be ignored").

Also tweaked the documentation comment to accurately describe the fileURL parameter; I tested and both an absolute path derived from an NSURL (currently in use) and a string path relative to the repository root work fine, so I didn't see any reason to change the method to match the documentation comment instead.


joshaber:
Ha, nice catch :+1:

Issue #541, opened 2016-01-26
[GTRepository shouldFileBeIgnored:success:error:] cannot be used from Swift


onecrayon:
This is actually a general problem that affects a few different methods throughout the codebase. The problem is that Swift assumes that the BOOL return value indicates success or failure, and as a result it is impossible to get the return value for this method (and others like it). In order to fix this it would need to be revised to something along these lines (hopefully with better naming):

    // Returns YES on success or NO on failure
    - (BOOL)shouldFile:(NSURL *)fileURL beIgnored:(BOOL *)ignored error:(NSError **)error;

I'm happy to revise methods that have this problem as I run into them and submit pull requests, but I wasn't sure what would be the best method for handling the change, since it's nowhere near backwards compatible (add a deprecated message for the original method? Just leave it alone and add a wrapper that will work properly in Swift? Also, how should I revise the method name, since I'm not too happy with the pattern that I used above?).

Thoughts?

tiennou:
I suggest `- (BOOL)fileAtURL:(NSURL *)fileURL isIgnored:(BOOL*)ignored error:(NSError **)error` ? That mostly-mimics `-[NSFileManager fileExistsAtPath:]`.

joshaber:
Well that's annoying :disappointed: 

I *think* I prefer `-shouldFile:beIgnored:error:` since it puts the verb in the first part of the selector, but I don't feel terribly strongly.

pietbrauer:
:+1: for Josh's proposal

onecrayon:
I've been giving this some thought, and since the primary reason to introduce a new method at all is for use in Swift there's another pattern that might make sense:

    // Returns 1 if the file should be ignored, 0 should not be ignored, -1 for failure
    - (NSInteger)shouldIgnoreFileURL:(NSURL *)fileURL;

Or alternately use an enum:

    typedef NS_ENUM(NSInteger, GTFileIgnoreState) {
        GTFileIgnoreStateIgnoreCheckFailed = -1,
        GTFileIgnoreStateShouldNotIgnore = 0,
        GTFileIgnoreStateShouldIgnore = 1
    };
    - (GTFileIgnoreState)shouldIgnoreFileURL:(NSURL *)fileURL;

[...]

PR #545, opened 2016-01-28
Added shouldIgnoreFileURL:error: convenience wrapper

onecrayon:
Added to GTRepository+Status for compatibility with Swift; closes issue #541.

### DIFF
--- a/ObjectiveGit/GTRepository+Status.m
+++ b/ObjectiveGit/GTRepository+Status.m
@@ -125,4 +125,10 @@ NSString *const GTRepositoryStatusOptionsPathSpecArrayKey = @"GTRepositoryStatus
 	return (ignoreState == 1 ? YES : NO);
 }
 
+- (GTFileIgnoreState)shouldIgnoreFileURL:(NSURL *)fileURL error:(NSError **)error {
+	BOOL success = false;
+	BOOL ignore = [self shouldFileBeIgnored:fileURL success:&success error:error];
+	return ignore && success ? GTFileIgnoreStateShouldIgnore : (!ignore && success ? GTFileIgnoreStateShouldNotIgnore : GTFileIgnoreStateIgnoreCheckFailed);
+}
+
 @end
--- a/ObjectiveGit/GTRepository+Status.m
+++ b/ObjectiveGit/GTRepository+Status.m
@@ -126,9 +126,12 @@ NSString *const GTRepositoryStatusOptionsPathSpecArrayKey = @"GTRepositoryStatus
 }
 
 - (GTFileIgnoreState)shouldIgnoreFileURL:(NSURL *)fileURL error:(NSError **)error {
-	BOOL success = false;
+	BOOL success = NO;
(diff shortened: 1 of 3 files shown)


pietbrauer:
For my feeling this is a lot of bool comparison to go right the first time. Could you add a test for the new method?

ObjectiveGit/GTRepository+Status.m
```diff
@@ -125,4 +125,10 @@ - (BOOL)shouldFileBeIgnored:(NSURL *)fileURL success:(BOOL *)success error:(NSEr
 	return (ignoreState == 1 ? YES : NO);
 }
 
+- (GTFileIgnoreState)shouldIgnoreFileURL:(NSURL *)fileURL error:(NSError **)error {
+	BOOL success = false;
```
joshaber: Conventionally this should be `NO` instead of `false`.

ObjectiveGit/GTRepository+Status.m
```diff
@@ -125,4 +125,10 @@ - (BOOL)shouldFileBeIgnored:(NSURL *)fileURL success:(BOOL *)success error:(NSEr
...
+	BOOL ignore = [self shouldFileBeIgnored:fileURL success:&success error:error];
+	return ignore && success ? GTFileIgnoreStateShouldIgnore : (!ignore && success ? GTFileIgnoreStateShouldNotIgnore : GTFileIgnoreStateIgnoreCheckFailed);
```
joshaber: This is a bit complex to squeeze into one ternary. How about splitting them up?

joshaber:
Thanks @onecrayon! This is close, just a couple comments.

>Could you add a test for the new method?

Yup, a test or two would be great :+1:

ObjectiveGit/GTRepository+Status.m
onecrayon: Ha; guess it shows that I've been coding nothing but Swift for the past while...

onecrayon:
I am not familiar with the testing framework being used here, and there does not appear to be any test associated with the method this is wrapping, so mimicking that is out (e.g. -statusForFile:success:error). I've been testing this locally, and the logic is fine (you'll note my previous pull request fixing the broken logic in the method I'm wrapping...), but if you still want a test added to the suite you'll need to point me in the direction of a good example that I can customize for this particular method.

pietbrauer:
True, the existing implementation is not being tested directly. There are some status specs [here](https://github.com/libgit2/objective-git/blob/master/ObjectiveGitTests/GTRepository%2BStatusSpec.m#L96) you could just copy paste one if them and check if your method returns the right enum value.

onecrayon:
Added a couple of tests: one for the original method and one for the wrapper; both are passing for me. Although FYI the testing suite as a whole is crashing and burning on a memory error for an unrelated test later on:

    -[GTSignatureSpec should_keep_the_git_signature_alive_even_if_the_object_goes_out_of_scope]

joshaber:
Looks good, thanks!

PR #484, opened 2015-07-10
iOS Test target

phatblat:
iOS Tests #444 

screen shot 2015-07-10 at 1 25 16 am

To help prevent things like #478 & #481.

Using [SSZipArchive](https://github.com/ZipArchive/ZipArchive) isn't ideal since it isn't a framework so the Xcode integration is ugly. But at least we can now run tests on iOS. Boy, are they slow. :zzz:

PR #552, opened 2016-02-16
Fix fallback to iphonesimulator in script/cibuild

tombooth:
In https://github.com/libgit2/objective-git/pull/484 more flags were
added into the variable `sdkflag`, which then caused the following
execution of run_xctool to always fail. This means that iOS tests have
not been running as part of the TravisCI run, they have only been built.
You can see this if you look at the final run for the PR:
https://travis-ci.org/libgit2/objective-git/builds/78334551#L192

Bash does some great things when trying to interpret variables in
commands. Initially the `$sdkflag` didn't have any quotes around it as
can be seen in commit 15f906ca2efa9d5ba67744d25a825c9bc75ad9e7. When the
second flag was added this would have caused the command to fail because
of the way Bash would interpret the whitespace. Without the surrounding
quotes the command would have been executed as follows:

```
xctool -workspace ObjectiveGitFramework.xcworkspace RUN_CLANG_STATIC_ANALYZER=NO -sdk iphonesimulator -destination '"platform=iOS' Simulator,name=iPhone '5"' -scheme "ObjectiveGit iOS" test ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO
```

You can see in the above that the white space was split in the middle of
the `-destination` parameter and it causes `xctool` to spout out an
error. When the surrounding quotes were added it caused xctool to run
but `$sdkflag` would be interpretted as a single argument, as it is
executed as follows:

```
xctool -workspace ObjectiveGitFramework.xcworkspace RUN_CLANG_STATIC_ANALYZER=NO '-sdk iphonesimulator -destination "platform=iOS Simulator,name=iPhone 5"' -scheme "ObjectiveGit iOS" test ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO
```

Notice in the above that the entire argument is still surrounded by
single quotes.

This can be solved by switching `$sdkflag` for an array of strings, this
can then be correctly expanded in the command getting around the issues
highlighted above.

You can read more about these issues with Bash in the following links:

  - http://mywiki.wooledge.org/BashFAQ/050
  - http://mywiki.wooledge.org/BashFAQ/073

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -118,7 +118,7 @@ build_scheme ()
     echo "*** Building and testing $scheme..."
     echo
 
-    local sdkflag=
+    local sdkflags=()
     local action=test
 
     # Determine whether we can run unit tests for this target.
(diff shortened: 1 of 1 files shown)


tombooth:
Unsure why it isn't building, but you can see the results of a run here: https://travis-ci.org/tombooth/objective-git/builds/109612507

pietbrauer:
Hah, tricked Travis :muscle:

pietbrauer:
Nice catch, looks good to me. All iOS and Mac Tests run through. It will take a bit longer now (40 vs. 24 minutes), but will lead to better coverage of course.

PR #554, opened 2016-02-24
Add matrix build for iOS and Mac

pietbrauer:
If this works then the time builds will take is only half as we split the build for Mac and iOS to separate machines.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
-osx_image: xcode7
+osx_image: xcode7.2
 language: objective-c
+matrix:
+  include:
+    - osx_image: xcode7.2
+      env: SCHEME="ObjectiveGit Mac"
+    - osx_image: xcode7.2
+      env: SCHEME="ObjectiveGit iOS"
 before_install:
   - brew update
   - brew outdated xctool || brew upgrade xctool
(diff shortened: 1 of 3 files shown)


phatblat:
```
../SwiftSpec.swift:11:8: error: module file's minimum deployment target is ios9.2 v9.2: /Users/travis/Library/Developer/Xcode/DerivedData/ObjectiveGitFramework-dggdbcqvjtimfddtjpjrzodjcbls/Build/Products/Debug-iphonesimulator/Quick.framework/Modules/Quick.swiftmodule/x86_64.swiftmodule
import Quick
```

This is an odd error because there are no targets that I can find that have a deployment target of iOS 9.2. Based on the SO post below, I suspect this is might be related to the iOS Tests target being raised to `9.0`.

http://stackoverflow.com/questions/31461012/can-no-longer-import-app-module-into-unit-tests-after-changing-deployment-target

I can't reproduce this locally so can't tell if this would resolve the issue. Why was the Test deployment target raised?

pietbrauer:
I think I did that to fix other errors regarding the Quick update which now only works with iOS 9.0? Will try to dig into it today.

pietbrauer:
@phatblat Green now, seems to be related to Xcode 7.2

phatblat:
Works for me

PR #555, opened 2016-02-24
Fix super slow iOS Tests

pietbrauer:
Updates dependencies to Nimble, Quick and SSZipArchive.

iOS Tests were running super slow (30 minutes total) this was because after every single test case the unzipped fixture repository would be deleted from disk and unzipped again from the 28MB big zip file. This was not an issue on Mac as `unzip` super fast (apparently) but `SSZipArchive` cannot quite keep up and so every single iOS test case would take 5-7 seconds to unzip + the time the test needs to run.

What I did is just unzip it once and store it in a directory called `clean_repository`, the directory is still removed after each test but instead of unzipping again I just copy the desired repository into its place not touching the `clean_repository`.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,4 +1,4 @@
 github "jspahrsummers/xcconfigs" >= 0.7.1
 github "Quick/Quick" ~> 0.9.1
 github "Quick/Nimble" ~> 3.1
-github "ZipArchive/ZipArchive" ~> 0.3
+github "ZipArchive/ZipArchive" == 1.0.1
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,4 +1,4 @@
 github "jspahrsummers/xcconfigs" >= 0.7.1
-github "Quick/Quick" ~> 0.6
-github "Quick/Nimble" ~> 2.0
+github "Quick/Quick" ~> 0.9.1
+github "Quick/Nimble" ~> 3.1
 github "ZipArchive/ZipArchive" ~> 0.3
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-github "Quick/Nimble" "v2.0.0"
-github "Quick/Quick" "v0.6.0"
+github "Quick/Nimble" "v3.1.0"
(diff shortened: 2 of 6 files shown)


pietbrauer:
ITS GREEEN!!!! 



Due to the Quick Update I had to set the deployment target of the Tests to iOS 9.0 which does not affect the framework itself.

A nice side effect of the Quick Update is that test names are now actually readable. :v:

phatblat:
This is amazingly faster. Nice work!

Had one snag when running `carthage bootstrap`:

```
A shell task failed with exit code 74:
xcodebuild: error: Unable to read project 'SSZipArchive.xcodeproj' from folder '/Users/phatblat/dev/libgit2/objective-git/Carthage/Checkouts/ZipArchive'.
	Reason: Project /Users/phatblat/dev/libgit2/objective-git/Carthage/Checkouts/ZipArchive/SSZipArchive.xcodeproj cannot be opened because it is missing its project.pbxproj file.
```

It looks like checking out the [`v1.0.1`](https://github.com/ZipArchive/ZipArchive/tree/v1.0.1) tag left an empty `SSZipArchive.xcodeproj` folder since there is no `project.pbxproj` in the tree at that tag.

This was due to having an older ZipArchive checkout which left behind the `.xcodeproj` file and confused carthage and xcodebuild.

phatblat:
One little nitpick, there's now a dead workspace reference to the SSZipArchive project.

screen shot 2016-02-25 at 8 26 33 pm

pietbrauer:
Yeah, SSZipArchive is a bit weird but we use the files directly.

pietbrauer:
Will remove the file reference.

phatblat:
Looks like [`v1.1`](https://github.com/ZipArchive/ZipArchive/tree/v1.1) has framework targets now, but that can be updated in another PR.

pietbrauer:
Updated SSZipArchive `v1.1` and fixed the reference.

pietbrauer:
Can't update to 1.1 because it has a lot of warnings. Staying on 1.0.1 but remove the reference to the project.

phatblat:
Sounds good to me!

PR #559, opened 2016-03-02
Write merge conflicted files when pulling

pietbrauer:
This mimics the default `git` behaviour and writes the conflicted files down like this:

```
<<<<<HEAD
fksdjfa;lsdkjf;lads;afl
========
<<<<<<<<<4567dvssqw3124323
```

It also writes into `.git/MERGE_HEAD` the remote commit SHA, into `.git/MERGE_MSG` the correct merge message. It can then be pulled in the merge commit via `try repository.preparedMessage()`.

The merge head enumeration was added so the merge commit can pull all merge heads and use them as parent commits.

### DIFF
--- a/ObjectiveGit/GTRepository+Merging.h
+++ b/ObjectiveGit/GTRepository+Merging.h
@@ -45,11 +45,10 @@ typedef NS_OPTIONS(NSInteger, GTMergeAnalysis) {
 
 /// Merge Branch into current branch
 ///
-/// analysis   - The resulting analysis.
 /// fromBranch - The branch to merge from.
 /// error      - The error if one occurred. Can be NULL.
 ///
-/// Returns YES if the nerge was successful, NO otherwise (and `error`, if provided,
+/// Returns YES if the merge was successful, NO otherwise (and `error`, if provided,
 /// will point to an error describing what happened).
 - (BOOL)mergeBranchIntoCurrentBranch:(GTBranch *)fromBranch withError:(NSError **)error;
 
--- a/ObjectiveGit/GTRepository+Merging.h
+++ b/ObjectiveGit/GTRepository+Merging.h
@@ -1,0 +1,34 @@
+//
+//  GTRepository+Merging.h
+//  ObjectiveGitFramework
+//
+//  Created by Piet Brauer on 02/03/16.
+//  Copyright © 2016 GitHub, Inc. All rights reserved.
+//
+
+#import "GTRepository.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GTRepository (Merging)
+
+/// Enumerate all available merge head entries.
+///
+/// error - The error if one ocurred. Can be NULL.
+/// block - A block to execute for each MERGE_HEAD entry. `mergeHeadEntry` will
+///         be the current merge head entry. Setting `stop` to YES will cause
+///         enumeration to stop after the block returns. Must not be nil.
+///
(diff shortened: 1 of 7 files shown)


pietbrauer:
This has tests now :white_check_mark:

pietbrauer:
I revised this PR and move the merge algorithm to it's own function so it can be used independently when merging branches.

The reason I did this was because I was beginning to implement merging branches in Git2Go and saw that is was exactly the same procedure as pulling and we should probably factor it out. This also made pulling way easier to grasp.

I am open to suggestions regarding the naming of the merge function.

ObjectiveGit/GTRepository+Merging.h
```diff
@@ -0,0 +1,68 @@
...
+/// Convenience method for -enumerateMergeHeadEntriesWithError:usingBlock: that retuns an NSArray with all the fetch head entries.
+///
+/// error - The error if one ocurred. Can be NULL.
+///
+/// Retruns a (possibly empty) array with GTCommit objects. Will not be nil.
+- (NSArray <GTCommit *>*)mergeHeadEntriesWithError:(NSError **)error;
+
+/// Merge Branch into current branch
+///
+/// analysis   - The resulting analysis.
```
joshaber: Looks like this isn't an argument anymore.

ObjectiveGit/GTRepository+Merging.h
```diff
@@ -0,0 +1,68 @@
...
+/// fromBranch - The branch to merge from.
+/// error      - The error if one occurred. Can be NULL.
+///
+/// Returns YES if the nerge was successful, NO otherwise (and `error`, if provided,
```
joshaber: s/nerge/merge :wine_glass:

ObjectiveGit/GTRepository+Merging.m
```diff
@@ -0,0 +1,215 @@
...
+	__unsafe_unretained GTRepositoryEnumerateMergeHeadEntryBlock enumerationBlock;
+} GTEnumerateMergeHeadEntriesPayload;
+
+int GTMergeHeadEntriesCallback(const git_oid *oid, void *payload) {
+	GTEnumerateMergeHeadEntriesPayload *entriesPayload = payload;
+
+	GTRepository *repository = entriesPayload->repository;
+	GTRepositoryEnumerateMergeHeadEntryBlock enumerationBlock = entriesPayload->enumerationBlock;
+
+	GTCommit *commit = [repository lookUpObjectByOID:[GTOID oidWithGitOid:oid] objectType:GTObjectTypeCommit error:NULL];
```
joshaber: I wonder if the merge head methods should yield OIDs instead of commits, so users don't incur the lookup overhead unless they actually need the commits. Thoughts?

joshaber:
:dancer: Just the one question left.

ObjectiveGit/GTRepository+Merging.m
pietbrauer: Hm, you basically take the results of this method and put it into the parents of the merge commit. I mean it's only a matter of another map so I omitted that. But we totally could switch to `GTOID`s as this would map 1:1 to the libgit2 functionality.

pietbrauer:
@joshaber Changed the implementation to return GTOID objects :rocket:

pietbrauer:
Would be nice to have a new version including this and Bitcode :v:

joshaber:
:metal:

PR #562, opened 2016-03-17
Ignore some more directories

Uncommon:
Adds the following to the ignore list:
- ObjectiveGitTests/expecta/
- ObjectiveGitTests/specta/
- External/Configuration

These untracked directories are causing objective-git to show as a "dirty" submodule for parent repositories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,11 @@ project.xcworkspace
 *.xccheckout
 *.xcscmblueprint
 
+ObjectiveGitTests/expecta/
 ObjectiveGitTests/fixtures/Fixtures/*
+ObjectiveGitTests/specta/
 
+External/Configuration
 External/ios-openssl/include
 External/ios-openssl/lib
 External/libssh2-ios


pietbrauer:
I wonder who creates these directories? They don't appear when I work on ObjectiveGit only.

Uncommon:
I did a quick search through the build scripts but couldn't find where they come from.

pietbrauer:
I would like to find out why it happens in your submodule and not when I work on it directly or using another project before merging this.

pietbrauer:
Just curious, are you using specta and expecta in your project? Because objective-Git uses Nimble and Quick for Tests.

Uncommon:
I think I know what happened. Those directories may be left over from an older version of objective git, and when I checked out the latest they didn't get deleted. I'll double check and close this if I'm right.

Uncommon:
Looks like I was right. I removed the directories, rebuilt, and they didn't reappear. Pull request retracted :)

PR #564, opened 2016-03-22
Update libgit2 to 0.24.0

pietbrauer:
This updates `libgit2` to the latest release `v0.24.0`:

~~I had to disable `GIT_OPT_ENABLE_STRICT_OBJECT_CREATION` in these tests.~~

* ~~`[GTRemotePushSpec pushing__to_remote__can_push_one_commit]`~~
* ~~`[GTRemoteSpec network_operations____GTRepository_fetchRemote_withOptions_error_progress____brings_in_new_commits]`~~
* ~~`[GTRepositoryCommitting can_create_commits]`~~
* ~~`[GTRepositoryPullSpec pull__from_remote__fails_to_merge_when_there_is_a_conflict]`~~
* ~~`[GTTreeBuilderSpec GTTreeBuilder_building__should_write_new_blobs_when_the_tree_is_written]`~~

~~Now I am unsure if I should stick to [permanently disabling this](https://github.com/libgit2/objective-git/commit/c51e7387c4d53a1cc6dd44ad36338c9b98e8c1c3) or [just silence it for the tests](https://github.com/libgit2/objective-git/commit/cfe733ba32eb1df96eaedf822d1b63b160c1567d).~~

### DIFF
--- a/ObjectiveGit/GTRepository+RemoteOperations.m
+++ b/ObjectiveGit/GTRepository+RemoteOperations.m
@@ -226,7 +226,7 @@ int GTFetchHeadEntriesCallback(const char *ref_name, const char *remote_url, con
 	remote_callbacks.push_transfer_progress = GTRemotePushTransferProgressCallback;
 	remote_callbacks.payload = &connectionInfo,
 
-	gitError = git_remote_connect(remote.git_remote, GIT_DIRECTION_PUSH, &remote_callbacks);
+	gitError = git_remote_connect(remote.git_remote, GIT_DIRECTION_PUSH, &remote_callbacks, NULL);
 	if (gitError != GIT_OK) {
 		if (error != NULL) *error = [NSError git_errorFor:gitError description:@"Failed to connect remote"];
 		return NO;


pietbrauer:
This is where it was added: https://github.com/libgit2/libgit2/pull/3633

I am unsure why our additions to the tree are considered invalid. The `OID` is generated using `+[GTOID OIDByHashingData:type:error:]`, file mode is valid is set to `GTFileModeBlob` and the file name is valid too.

Maybe @ethomson has an idea?

pietbrauer:
Ready for review, once this is merged I will do a new release.

ethomson:
Well, for `[GTTreeBuilderSpec GTTreeBuilder_building__should_write_new_blobs_when_the_tree_is_written]`, for example, the process is:

1. Hash the blob (in `[GTTreeBuilder addEntryWithData]`)
2. Add to the pending data list
3. Add the tree entry with the blob's ID / filename / mode, by calling `git_treebuilder_insert`

You're not actually writing the blob, though, until `[GTTreeBuilder writeTree]`, which will run `[GTTreeBuilder writePendingData]` and actually flush the pending list to blobs.

When you call `git_treebuilder_insert` (in step 3), it's going to check that a blob exists in the database.   It doesn't yet.  So I expect this failure.

I don't see why Objective C isn't simply adding the blob to the ODB in `[GTTreeBuilder addEntryWithData]`...?  You're spending a lot of memory and CPU to avoid doing so...  Unless there's a really good reason to avoid it, I would make `[GTTreeBuilder addEntryWithData]` add the given data to the ODB, then call `addEntryWithOID` with the new OID, and get rid of the pending data list entirely.

(Note, please correct my notation if I'm putting brackets in the wrong place, I'm very much a noob when it comes to Objective C.)

I only looked at this one test - it's possible that the other tests have the same root cause (or not).

I was hoping to actually step through these with a debugger, but I have failed to do so as of yet.  :(

Do you have or can you point me to a quick getting started for this, for somebody who is *not* an Objective C developer and maybe doesn't understand how frameworks work?  The README suggested that as long as I was running Xcode 7 that I would be set, so I ran `script/bootstrap`, but my ObjectiveGit-MacTests build fails to include the frameworks.  (`SwiftSpec.swift: No such module 'Nimble'`).

I also tried building them with carthage and dragging the resultant frameworks in with no avail.

pietbrauer:
Oh, I think I have to reread this when I am more awake :laughing: 

You should be able to run the tests if all the submodules are there (including Nimble). Did you maybe open the `.xcodeproj` instead of the `.xcworkspace`?

ethomson:
Yes, that's exactly what I did.  Thanks for the tip.  :)

pietbrauer:
Reworked this to include the changes made in #566 so the tests pass now. Once this is merged I will release `0.12.0` and we can all update to Xcode 7.3 :grin:

/cc @joshaber @phatblat

phatblat:
Looks good to me.

phatblat:
At least Nimble will need an update to 3.2 for basic Swift 2.2/Xcode 7.3 compatibility. However, there are a ton of deprecation warnings that I've only just fixed tonight in Quick and Nimble. I'll submit a PR to update these to the commits in master since they haven't been included in a release yet.

pietbrauer:
Cool, but this is not blocking the release right?

phatblat:
No. It's just blocking running the tests on Xcode 7.3. I had Travis failures from building and testing Quick due to the spurious warning output. Apparently if the build writes more than 4MB of logs it's considered a failure.

pietbrauer:
Fun fact: it is blocking the release as Carthage builds Nimble & Quick. :laughing:

PR #566, opened 2016-03-23
GTTreeBuilder: don't add blob data lazily

ethomson:
Add the blob data to the object database immediately in `[GTTreeBuilder addEntryWithData]`, instead of queueing it up.  This enables strict object validity checking, prevents us from hashing the file twice (once to compute the OID to give to `addEntryWithOID`, and again when adding to the object database) and reduces memory overhead.

### DIFF
--- a/ObjectiveGit/GTTreeBuilder.m
+++ b/ObjectiveGit/GTTreeBuilder.m
@@ -43,13 +43,6 @@
 @property (nonatomic, assign, readonly) git_treebuilder *git_treebuilder;
 @property (nonatomic, strong, readonly) GTRepository *repository;
 
-// Data to be written with the tree, keyed by the file name. This should only be
-// accessed while synchronized on self.
-//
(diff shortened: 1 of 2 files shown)


ethomson:
Note, of course, that I'm missing the historical context here.  So I may be missing the use case that this enables.  I think that in the general case, removing this is probably the right thing, but if there's value here then it could be optional instead...

pietbrauer:
Seem like the initial implementation was made back in 2013 (https://github.com/libgit2/objective-git/commit/1f203eb58015923bddfb5d5a71226c080f99b80f) but it seems to have no specific reason to write the data to the `odb` at once.

This also fixes the test for the `0.24.0` which is stricter with writing the correct objects, so my guess is: Because libgit2 did not complain back then it was ok to do it like that. Now that it is complaining we should think of the proper way of doing it, which is this implementation.

Thanks so much for sorting this out! :sparkles:

PR #575, opened 2016-05-19
Support for Git Notes

slavikus:
Adding support for Git Notes already supported in libgit2. A new class, `GTNote` has been added, along with a few methods for `GTRepository` to create and remove notes.


slavikus:
Sorry, going to re-submit with push of notes support.

PR #577, opened 2016-05-24
Update the OS X deployment target from 10.8 to 10.11

Uncommon:
Every time I build my project, I get a slew of warnings because libgit2 was built for 10.11, but ObjGit is targeting 10.8.

### DIFF
--- a/ObjectiveGitFramework.xcodeproj/project.pbxproj
+++ b/ObjectiveGitFramework.xcodeproj/project.pbxproj
@@ -1706,7 +1706,7 @@
 					.,
 					External,
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.8;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				ONLY_ACTIVE_ARCH = YES;
 				OTHER_CFLAGS = (
 					"$(inherited)",
@@ -1736,7 +1736,7 @@
 					.,
 					External,
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.8;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-DGIT_SSH",
@@ -1856,7 +1856,7 @@
 					.,
 					External,
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.8;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-DGIT_SSH",
@@ -2089,7 +2089,7 @@
 					.,
 					External,
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.8;
+				MACOSX_DEPLOYMENT_TARGET = 10.11;
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-DGIT_SSH",


Uncommon:
The Travis build failure is on the iOS target, which I didn't change. I don't know what's going on there.

pietbrauer:
I did some digging around, because it annoyed me too. It seems like this is a bug in Xcode and should not cause a warning. I am strongly against setting the deployment target up, as it breaks other apps. GitHub Desktop for example is targeting 10.9.

Please see this thread: https://forums.developer.apple.com/thread/20024

Uncommon:
OK, sounds like this isn't the right solution then. Though I'm left wondering why libgit2 is getting built for 10.11. Maybe the build script doesn't specify, so it defaults to the latest.

phatblat:
@Uncommon Are the warnings coming from the linker phase of the libgit2 target?

```
ld: warning: object file (/usr/local/lib/libssh2.a(agent.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(channel.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(hostkey.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(session.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(userauth.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(dsa_lib.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(digest.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(rand_lib.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(rsa_lib.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(misc.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(openssl.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(global.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(kex.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(packet.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(transport.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (/usr/local/lib/libssh2.a(keepalive.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(asn1_lib.o)) was built for newer OSX version (10.10) than being linked (10.8)
ld: warning: object file (External/libcrypto.a(bio_lib.o)) was built for newer OSX version (10.10) than being linked (10.8)
```

These say 10.10 because I captured them last year.

phatblat:
Based on [this SO answer](http://stackoverflow.com/questions/10165335/can-cmake-specify-the-base-sdk-on-mac-os-x#answer-10559847), the fix may be to specify the `CMAKE_OSX_SYSROOT` like we do in [`update_libgit2_ios`](https://github.com/libgit2/objective-git/blob/master/script/update_libgit2_ios#L34)

phatblat:
Note that #570 resolves these warnings because libgit2 is built from an Xcode project instead of cmake.

Uncommon:
The warnings show up under the ObjectiveGit-Mac target rather than libgit2, but they look just like that.

pietbrauer:
@Uncommon do you still get these errors in Xcode 8?

Uncommon:
Yes, I still see them in Xcode 8 beta 2.

Uncommon:
Also in Xcode 8 final.

tiennou:
The libgit2 link might be fixable (we're in control of that, we might just need to pass the correct MACOSX_DEPLOYMENT_TARGET through [CMake](https://cmake.org/cmake/help/v3.0/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html). libssh2 isn't so easy — AFAIK there's no way to ask homebrew to do that...

PR #580, opened 2016-06-27
Make GTCredential from GTCredentialProviderBlock optional

alehed:
Currently the credential is enforced as non-optional by the swift compiler.
This means that there is no way for the user to abort a remote operation that demands a credential.


phatblat:
We've only been adding the _Nullable annotations to the .h files as that's the only place I think they are required. Not sure which is the better practice. 

I like that you use the newer _Nullable. It's probably time to change the rest of the uses of the old nullable annotation, but that can be done in another PR.

alehed:
Ok, that's fixed. Any ideas to why the tests are failing?

phatblat:
Nice. Not sure what happened with that one build ¯\_(ツ)_/¯

PR #582, opened 2016-06-28
New nullable annotation

alehed:
While I'm at it. Also removes some nullability annotations from the implementation files. 
Depends on #580 to be merged first.

### DIFF
--- a/ObjectiveGit/GTBlame.h
+++ b/ObjectiveGit/GTBlame.h
@@ -24,7 +24,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// blame - A git_blame to wrap. May not be NULL.
 ///
 /// Returns a blame, or nil if initialization failed.
-- (nullable instancetype)initWithGitBlame:(git_blame *)blame NS_DESIGNATED_INITIALIZER;
+- (instancetype _Nullable)initWithGitBlame:(git_blame *)blame NS_DESIGNATED_INITIALIZER;
 
 /// Get all the hunks in the blame. A convenience wrapper around `enumerateHunksUsingBlock:`
 @property (nonatomic, strong, readonly) NSArray<GTBlameHunk *> *hunks;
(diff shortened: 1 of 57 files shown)


tiennou:
Any specific reason for the change (other that it's new in Xcode 7) ? I find that painful-er to read, and I haven't been able to find what's the current preferred syntax (Apple seems to use the non-underscored variant unless there's wacky pointers-to-pointers).

[This](https://developer.apple.com/swift/blog/?id=25) and [this](http://clang.llvm.org/docs/AttributeReference.html#nullability-attributes) seem to imply that `_Nullable` was added for compatibility with C code that already used `__nullable`, and that `nullable` is only available to Obj-C.

Just my 2 cents, in any case.

alehed:
The main argument is consistency. It avoids the use of both __nullable, nullable and possibly _Nullable in the same file. Also it says in the blog post that in Xcode 7 __nullable is now just a macro that expands to _Nullable. 

As to the aesthetic consideration, I agree the property syntax with nullable looks nicer but in the other cases I think the older syntax is not justified and also this is Obj-C we're talking about.

Anyway, @phatblat suggested it so I thought it would be a good time to update the syntax.

alehed:
If you think it is cleaner, I can revert it in the `@property` cases. In the other cases I think `_Nullable` is the way to go.

pietbrauer:
@alehed Thanks for the PullReuqest. I also don't like it but it [seems this is the way to go](https://developer.apple.com/swift/blog/?id=25) in Xcode 7 and above. Could you rebase onto master so we can merge this?

alehed:
Should I leave the ones with `@property` syntax alone or should everything be `_Nullable`?

pietbrauer:
I would go with the new Xcode way and use `_Nullable`

tiennou:
Those `_nullable` annotations are for compatibility with someones' else C attribute that was found to be clashing.

From the Nullability and Objective-C document linked above, emphasis mine :

- *You can use _Nullable and _Nonnull almost anywhere you can use the normal C const keyword, though of course they have to apply to a pointer type. **However, in the common case there’s a much nicer way to write these annotations**: within method declarations you can use the non-underscored forms nullable and nonnull immediately after an open parenthesis, as long as the type is a simple object or block pointer.*

- *This feature was first released in Xcode 6.3 with the keywords __nullable and __nonnull. **Due to potential conflicts with third-party libraries, we’ve changed them in Xcode 7** to the _Nullable and _Nonnull you see here.*

alehed:
Rebased on top of master.

pietbrauer:
Thanks a lot for your contribution!

PR #589, opened 2016-09-15
Add some nullables to methods & properties that can return nil

Uncommon:
Xcode 8 now complains about methods and properties that return nil but are not declared `nullable`. This fixes all of those analyzer warnings.

### DIFF
--- a/ObjectiveGit/GTCredential.h
+++ b/ObjectiveGit/GTCredential.h
@@ -46,7 +46,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// type     - the credential types allowed by the operation.
 /// URL      - the URL the operation is authenticating against.
 /// userName - the user name provided by the operation. Can be nil, and might be ignored.
-- (GTCredential * _Nullable)credentialForType:(GTCredentialType)type URL:(NSString *)URL userName:(nullable NSString *)userName;
+- (GTCredential * _Nullable)credentialForType:(GTCredentialType)type URL:(nullable NSString *)URL userName:(nullable NSString *)userName;
 @end
 
 /// The GTCredential class is used to provide authentication data.
--- a/ObjectiveGit/GTOID.m
+++ b/ObjectiveGit/GTOID.m
@@ -29,8 +29,8 @@
 - (NSString *)SHA {
 	char *SHA = git_oid_tostr_s(self.git_oid);
 	NSString *str = [[NSString alloc] initWithBytes:SHA
-																					 length:GIT_OID_HEXSZ
-																				 encoding:NSUTF8StringEncoding];
+                                             length:GIT_OID_HEXSZ
+                                           encoding:NSUTF8StringEncoding];
 	NSAssert(str != nil, @"Failed to create SHA string");
 	return str;
 }
--- a/ObjectiveGit/GTOID.m
+++ b/ObjectiveGit/GTOID.m
@@ -27,13 +27,11 @@
 }
 
 - (NSString *)SHA {
-	char *SHA = malloc(GIT_OID_HEXSZ);
-	NSAssert(SHA != NULL, @"Failed to malloc SHA string");
-
-	git_oid_fmt(SHA, self.git_oid);
-
-	NSString *str = [[NSString alloc] initWithBytesNoCopy:SHA length:GIT_OID_HEXSZ encoding:NSUTF8StringEncoding freeWhenDone:YES];
-	NSAssert(str != nil, @"Failed to allocate SHA string");
+	char *SHA = git_oid_tostr_s(self.git_oid);
+	NSString *str = [[NSString alloc] initWithBytes:SHA
+																					 length:GIT_OID_HEXSZ
+																				 encoding:NSUTF8StringEncoding];
+	NSAssert(str != nil, @"Failed to create SHA string");
 	return str;
 }
 
--- a/ObjectiveGit/GTOID.m
+++ b/ObjectiveGit/GTOID.m
@@ -28,12 +28,12 @@
 
 - (NSString *)SHA {
 	char *SHA = malloc(GIT_OID_HEXSZ);
-	if (SHA == NULL) return nil;
+	NSAssert(SHA != NULL, @"Failed to malloc SHA string");
 
 	git_oid_fmt(SHA, self.git_oid);
 
 	NSString *str = [[NSString alloc] initWithBytesNoCopy:SHA length:GIT_OID_HEXSZ encoding:NSUTF8StringEncoding freeWhenDone:YES];
-	if (str == nil) free(SHA);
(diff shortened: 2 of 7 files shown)


ObjectiveGit/GTCredential.h
```diff
@@ -46,7 +46,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// type     - the credential types allowed by the operation.
 /// URL      - the URL the operation is authenticating against.
 /// userName - the user name provided by the operation. Can be nil, and might be ignored.
-- (GTCredential * _Nullable)credentialForType:(GTCredentialType)type URL:(NSString *)URL userName:(nullable NSString *)userName;
+- (GTCredential * _Nullable)credentialForType:(GTCredentialType)type URL:(nullable NSString *)URL userName:(nullable NSString *)userName;
```
Uncommon: `GTCredentialAcquireCallback` potentially calls this with a nil url.

tiennou: The libgit2 documentation implies that it's never NULL. At least `git_remote_connect` will return an error on a `NULL` URL. It seems the only other user is proxy support, and in that case it's not clear (there's a check skipping NULLs in `git_proxy_options_dup`).

tiennou:
Question : does NSAssert-ing that a return is not nil makes the warning go away ? Some of those changes look strange, eg. GTOID having a nullable `-SHA`, or GTReference and a nullable `-OID`...

Uncommon:
Yes it does. Would you like me to replace the null checks with `NSAssert`s?

tiennou:
Good question. From an API standpoint, I think asserting makes it cleaner. On the other hand, if it ever happens, we'll throw an exception at runtime, which is Not Cool.

Uncommon:
Let's think about the actual failure cases, then.

For GTOID.SHA:
- Failure to `malloc` 40 bytes. If that happens, you're in trouble anyway.
- `NSString.initWithBytesNoCopy` is declared as nullable, but doesn't specify when it might return null. Presumably it's if the given string doesn't fit the specified encoding. If `git_oid_fmt` yields something that's not UTF-8 compliant, then something is seriously wrong.

I think it's okay to assert in those cases. I'll proceed similarly with the others and see what you think.

ObjectiveGit/GTCredential.h
Uncommon: I can't even find where it's used, searching for `git_cred_acquire_cb` etc.

ObjectiveGit/GTReference.h
```diff
@@ -52,7 +52,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly, strong) GTRepository *repository;
 @property (nonatomic, readonly) GTReferenceType referenceType;
 @property (nonatomic, readonly) const git_oid *git_oid;
-@property (nonatomic, strong, readonly) GTOID *OID;
+@property (nonatomic, strong, readonly, nullable) GTOID *OID;
```
Uncommon: `git_reference_target` returns null if it's not a direct reference, so this should be nullable.

ObjectiveGit/GTSignature.h
```diff
@@ -42,10 +42,10 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly, copy, nullable) NSString *email;
 
 /// The time when the action happened.
-@property (nonatomic, readonly, strong) NSDate *time;
+@property (nonatomic, readonly, strong, nullable) NSDate *time;
```
Uncommon: The other properties are nullable, so this is consistent.

On the other hand, they all return nil if `git_signature` is nil, and it's not nullable, and it doesn't look like it should ever be null.

But since I'm making this consistent with the other properties, I'll leave that as a separate issue.

tiennou: IIRC it's done that way because libgit2 might fail to parse an object's headers (in some degenerate-odb cases), hence allowing some of them to be null.

ObjectiveGit/GTOID.m
```diff
@@ -28,12 +28,12 @@ - (const git_oid *)git_oid {
...
 - (NSString *)SHA {
 	char *SHA = malloc(GIT_OID_HEXSZ);
-	if (SHA == NULL) return nil;
+	NSAssert(SHA != NULL, @"Failed to malloc SHA string");
 
 	git_oid_fmt(SHA, self.git_oid);
 
 	NSString *str = [[NSString alloc] initWithBytesNoCopy:SHA length:GIT_OID_HEXSZ encoding:NSUTF8StringEncoding freeWhenDone:YES];
-	if (str == nil) free(SHA);
+	NSAssert(str != nil, @"Failed to allocate SHA string");
```
tiennou: Maybe

``` obj-c
  char *SHA = git_oid_tostr_s(self.git_oid); // Stored in TLS
  NSString *str = [[NSString alloc] initWithBytes:SHA length:GIT_OID_HEXSZ encoding:NSUTF8StringEncoding]; // So we copy the TLS data
  NSAssert(str != nil, @"Failed to allocate SHA string");
```

?

ObjectiveGit/GTCredential.h
tiennou: I used the regex `/->cred(entials|_acquire_cb)/` against libgit2 and got most of them.

Uncommon: It looks like it's probably supposed to be non-null, but to guarantee it I'd rather depend on someone who knows the code better.

ObjectiveGit/GTOID.m
Uncommon: Even better!

Uncommon:
I see the Travis failure, but it doesn't seem to be related to my changes.

Uncommon:
(bump) Any further changes needed on this?

ObjectiveGit/GTOID.m
```diff
@@ -27,13 +27,11 @@ - (const git_oid *)git_oid {
...
-	if (SHA == NULL) return nil;
-
-	git_oid_fmt(SHA, self.git_oid);
-
-	NSString *str = [[NSString alloc] initWithBytesNoCopy:SHA length:GIT_OID_HEXSZ encoding:NSUTF8StringEncoding freeWhenDone:YES];
-	if (str == nil) free(SHA);
+	char *SHA = git_oid_tostr_s(self.git_oid);
+	NSString *str = [[NSString alloc] initWithBytes:SHA
+																					 length:GIT_OID_HEXSZ
+																				 encoding:NSUTF8StringEncoding];
```
tiennou: Crazy indentation here 😉 .

Uncommon:
Another bump. Any more issues?

tiennou:
Not that I can see. A local run of `script/cibuild` worked here so I'm merging this. Thanks for the PR !

PR #590, opened 2016-09-22
Split the OID lookup from the object lookup in GTEnumerator

tiennou:
This is a first step toward handling shallow repositories, by separating the lookup done while enumerating.

Also provides the missing OID through the error.

Related to libgit2/libgit2#3058

### DIFF
--- a/ObjectiveGit/GTRepository.m
+++ b/ObjectiveGit/GTRepository.m
@@ -290,7 +290,10 @@ static int remoteCreate(git_remote **remote, git_repository *repo, const char *n
 		if (error != NULL) {
 			char oid_str[GIT_OID_HEXSZ+1];
 			git_oid_tostr(oid_str, sizeof(oid_str), oid);
-			*error = [NSError git_errorFor:gitError description:@"Failed to lookup object %s in repository.", oid_str];
+			*error = [NSError git_errorFor:gitError
+							   description:@"Failed to lookup object"
+								  userInfo:@{GTGitErrorOID: [GTOID oidWithGitOid:oid]}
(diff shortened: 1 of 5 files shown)


tiennou:
I would like a second pair of eyes on that one. IMHO it's a small change, and it allows us to not-unexpectedly-fail when working against shallow repositories.

ObjectiveGit/GTRepository.m
```diff
@@ -290,7 +290,10 @@ - (id)lookUpObjectByGitOid:(const git_oid *)oid objectType:(GTObjectType)type er
 		if (error != NULL) {
 			char oid_str[GIT_OID_HEXSZ+1];
 			git_oid_tostr(oid_str, sizeof(oid_str), oid);
-			*error = [NSError git_errorFor:gitError description:@"Failed to lookup object %s in repository.", oid_str];
+			*error = [NSError git_errorFor:gitError
+							   description:@"Failed to lookup object"
```
pietbrauer: Indentation seems a bit off

tiennou: It actually looks fine in Xcode, I've checked. I'd say GitHub gets confused because Holy Wars™ 😉 .

pietbrauer: You sure, Xcode is fine with mixing tabs and spaces, Git is not.

pietbrauer: You are right. Please remove the multi line error and make it one line. I looked up the other usages of this error method and it is a one liner everywhere. #StickingToTheStyleGuide

tiennou:
Fixed styling. I'll keep my fingers crossed that Travis doesn't hate me...

PR #593, opened 2016-10-08
Added git_repository_open_ext wrapper

glegrain:
Added a GTRepository convenience initializer to use git_repository_open_ext to permit "searching" from the given url.

### DIFF
--- a/ObjectiveGit/GTRepository.h
+++ b/ObjectiveGit/GTRepository.h
@@ -110,6 +110,16 @@ extern NSString * const GTRepositoryCloneOptionsCloneLocal;
 /// A NSURL pointing to a local file that contains PEM-encoded certificate chain.
 extern NSString *const GTRepositoryCloneOptionsServerCertificateURL;
 
+/// Repository extended open control flags for
+/// +initWithURL:flags:ceilingDirs:error:.
+///
+/// See respository.h for documentation of each individual flag.
+typedef NS_OPTIONS(UInt32, GTRepositoryOpenFlags) {
+	GTRepositoryOpenNoSearch = GIT_REPOSITORY_OPEN_NO_SEARCH,
+	GTRepositoryOpenCrossFS = GIT_REPOSITORY_OPEN_CROSS_FS,
+	GTRepositoryOpenBare = GIT_REPOSITORY_OPEN_BARE,
+};
+
 /// Initialization flags associated with `GTRepositoryInitOptionsFlags` for
 /// +initializeEmptyRepositoryAtFileURL:options:error:.
 ///
@@ -201,6 +211,18 @@ typedef NS_ENUM(NSInteger, GTRepositoryStateType) {
 /// Returns the initialized repository, or nil if an error occurred.
 + (nullable instancetype)repositoryWithURL:(NSURL *)localFileURL error:(NSError **)error;
 
+/// Convenience class initializer to find and open a repository with extended controls.
+///
+/// localFileURL - The file URL for the new repository. Cannot be nil.
+/// flags        - A combination of the `GTRepositoryOpenFlags` flags.
+/// ceilingDirs  - A GIT_PATH_LIST_SEPARATOR delimited list of path prefixes at
+///                which the search for a containing repository should terminate.
+///                Can be NULL.
+/// error        - The error if one occurs.
+///
+/// Returns the initialized repository, or nil if an error occurred.
++ (nullable instancetype)initWithURL:(NSURL *)localFileURL flags:(UInt32)flags ceilingDirs:(nullable const char *)ceilingDirs error:(NSError **)error;
+
 /// Convenience initializer which uses the default options.
 ///
 /// localFileURL - The file URL for the new repository. Cannot be nil.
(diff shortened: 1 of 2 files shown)


ObjectiveGit/GTRepository.h
```diff
@@ -110,6 +110,16 @@ extern NSString * const GTRepositoryCloneOptionsCloneLocal;
 /// A NSURL pointing to a local file that contains PEM-encoded certificate chain.
 extern NSString *const GTRepositoryCloneOptionsServerCertificateURL;
 
+/// Repository extended open control flags for
+/// +initWithURL:flags:ceilingDirs:error:.
+///
+/// See respository.h for documentation of each individual flag.
+typedef NS_OPTIONS(UInt32, GTRepositoryOpenFlags) {
```
tiennou: `NSInteger` instead of `UInt32`. I think `UInt32` comes from Carbon ?

ObjectiveGit/GTRepository.h
```diff
@@ -201,6 +211,18 @@ typedef NS_ENUM(NSInteger, GTRepositoryStateType) {
...
+///
+/// localFileURL - The file URL for the new repository. Cannot be nil.
+/// flags        - A combination of the `GTRepositoryOpenFlags` flags.
+/// ceilingDirs  - A GIT_PATH_LIST_SEPARATOR delimited list of path prefixes at
+///                which the search for a containing repository should terminate.
+///                Can be NULL.
+/// error        - The error if one occurs.
+///
+/// Returns the initialized repository, or nil if an error occurred.
++ (nullable instancetype)initWithURL:(NSURL *)localFileURL flags:(UInt32)flags ceilingDirs:(nullable const char *)ceilingDirs error:(NSError **)error;
```
tiennou: - `initWith...` class-method is confusing. It should be named `-repositoryWith...` instead.

- Use the typedef name instead of `UInt32`

ObjectiveGit/GTRepository.m
```diff
@@ -177,6 +181,22 @@ - (instancetype)initWithURL:(NSURL *)localFileURL error:(NSError **)error {
 	return [self initWithGitRepository:r];
 }
 
+- (instancetype)initWithURL:(NSURL *)localFileURL flags:(UInt32)flags ceilingDirs:(const char *)ceilingDirs error:(NSError **)error {
+	if (!localFileURL.isFileURL || localFileURL.path == nil) {
+		if (error != NULL) *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFileWriteUnsupportedSchemeError userInfo:@{ NSLocalizedDescriptionKey: NSLocalizedString(@"Invalid file path URL to initialize repository.", @"") }];
```
tiennou: `NSFileWriteUnsupportedScheme` ?

ObjectiveGit/GTRepository.h
```diff
@@ -209,6 +231,18 @@ typedef NS_ENUM(NSInteger, GTRepositoryStateType) {
...
+///
+/// localFileURL - The file URL for the new repository. Cannot be nil.
+/// flags        - A combination of the `GTRepositoryOpenFlags` flags.
+/// ceilingDirs  - A GIT_PATH_LIST_SEPARATOR delimited list of path prefixes at
+///                which the search for a containing repository should terminate.
+///                Can be NULL.
+/// error        - The error if one occurs.
+///
+/// Returns the initialized repository, or nil if an error occurred.
+- (nullable instancetype)initWithURL:(NSURL *)localFileURL flags:(UInt32)flags ceilingDirs:(nullable const char *)ceilingDirs error:(NSError **)error;
```
tiennou: Also, having both a class and an instance method is unwarranted IMHO. I'd drop this one.

ObjectiveGit/GTRepository.h
tiennou: - `const char *` isn't very Cocoa-y. What about passing an `NSArray <NSURL *> *` and doing the conversion internally ?

glegrain: - Others class-methods for `GTRepository` are using `initWithURL:...`. It make sense for me to stay consistent. (Plus, I don't know how `repositoryWithURL...` will translate into Swift.  I will experiment and see)

- `NSArray <NSURL *> *` it is 😃

ObjectiveGit/GTRepository.h
glegrain: Actually, is there any reason you would rather drop the instance initializer rather than the class initializer? Keeping the convenience instance initializer allows my Swift code to look cleaner: `GTRepository(url: url, flags: 0, ceilingDirs: nil)`

glegrain:
Any ideas how to fix Travis?

ObjectiveGit/GTRepository.h
tiennou: Usually, my Cocoa-fu says that convenience initializers get declared as class methods, designated initializers go under `-init...`. Arguably, this one could be kept as a designated initializer (since it ends up calling one of the handful `libgit2` functions that creates objects), but our designated initializer is actually `initWithGitRepository:`.

I'm not sure how that would like after being Swift-preprocessed though.

tiennou:
The iOS tests are flaky, so (unless you happen to find a solution 😉 ), you're better off just making sure the macOS run is green.

pietbrauer:
@glegrain Could you rebase onto master and see if the tests run through? #591 might have fixed the iOS tests, which now also run on Xcode 8

pietbrauer:
iOS is green, Mac is failing because of SSH check which is because of something else. So this should be good from a CO endpoint.

pietbrauer:
Thanks for the PR 👾

PR #598, opened 2016-11-23
Respecting nullability for various properties

slavikus:
Some properties were not marked as `nullable` while they could, in fact, return `nil`. Fixing that.


slavikus:
Apparently there's already an active pull request for this, so closing this one.

PR #601, opened 2016-12-05
GTObject GTIndexEntry initializers should be nullable

Uncommon:
I missed this one in my previous pull request.


tiennou:
Seems legit 😉

PR #605, opened 2016-12-18
Fix mergeBranchIntoCurrentBranch for conflicted state

slavikus:
`checkout_opts` passed to libgit2 missed a `GIT_CHECKOUT_SAFE` strategy, causing a dry run and only conflicted files being written out to the tree (losing all other non-conflicted ones).

This is because libgit2 has a safeguard check in `git_merge` call, auto-adding `GIT_CHECKOUT_SAFE` when no `checkout_opts` where specified, but since the Objective-Git code specified `GIT_CHECKOUT_ALLOW_CONFLICTS` flag, it wasn't applied, causing the behaviour outlined above:

`checkout_strategy = given_checkout_opts ?
        given_checkout_opts->checkout_strategy :
        GIT_CHECKOUT_SAFE;`


tiennou:
Thanks !

PR #609, opened 2017-01-05
Update to libgit2 0.25.0

Uncommon:
I'm eager to see this updated because of the revwalk fixes. The only significant API change seems to be in `git_remote_connect` with the new `proxy_opts` parameter. The sample code passes `NULL` so that's what I did.


pietbrauer:
Thanks!

PR #619, opened 2017-03-02
Revert to "default enabled" behavior of checking out when cloning

### LINKED ISSUE #618: GTRepositoryCloneOptionsCheckout are missing
@tiennou, we have lost `GTRepositoryCloneOptionsCheckout` constant — it's declared in the `GTRepository.h` file, but nowhere else, so if any third-party code attempts to use it, we get a link error saying it's not defined anywhere. Not sure if the relevant code is also gone or substituted by `GTCheckoutOptions` struct.

tiennou:
This adds back a renamed version of the old BOOL options and sets a GTCheckoutOptions with sane defaults.

Fixes #618

### DIFF
--- a/ObjectiveGit/GTRepository.h
+++ b/ObjectiveGit/GTRepository.h
@@ -71,10 +71,10 @@ extern NSString * const GTRepositoryCloneOptionsBare;
 
 /// An `NSNumber` wrapped `BOOL`, if NO, don't checkout the remote HEAD.
 /// Default value is `YES`.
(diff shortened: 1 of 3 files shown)


pietbrauer:
Thanks a lot!

tiennou:
Oops, forgot to rename the options in the tests. Will fix !

tiennou:
Should be good to go. Hopefully. I can't run tests locally, so this is a blind fix.

PR #621, opened 2017-03-09
Clarify GTDiff options argument documentation

nanotech:
The diff initializers refer to their options arguments taking "any of the above options", but both `GTDiffOptions` and `GTDiffFindOptions` are defined above them and the diff initializers really only accept `GTDiffOptions`.

I got this mixed up in practice, resulting [in this bugfix](https://github.com/gitx/gitx/pull/54/files).


pietbrauer:
Thanks a lot for the contribution! This makes perfect sense.

PR #622, opened 2017-03-16
GTIndex improvements and GTRepository+Status tweaks

slavikus:
(Credit for this pull request goes to @mcyril)

- Added GTIndex.checksum accessor
- Added addPathspecs/removePathspaces calls.
- Fixed proper nullability for enumerateConflictedFiles
- Converted GTRepositoryStatusFlags into NS_OPTIONS for proper handling from Swift
- Added missing flags for GTFileStatusFlags

### DIFF
--- a/ObjectiveGit/GTIndex.h
+++ b/ObjectiveGit/GTIndex.h
@@ -29,7 +29,9 @@
 
 #import <Foundation/Foundation.h>
 #include "git2/types.h"
+#include "git2/index.h"
 
+@class GTOID;
 @class GTIndexEntry;
 @class GTRepository;
 @class GTTree;
@@ -44,6 +46,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// The file URL for the index if it exists on disk; nil otherwise.
 @property (nonatomic, readonly, copy) NSURL * _Nullable fileURL;
+
+/// The index checksum
+@property (nonatomic, readonly, strong) GTOID * _Nullable checksum;
 
 /// The number of entries in the index.
 @property (nonatomic, readonly) NSUInteger entryCount;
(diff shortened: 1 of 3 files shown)


ObjectiveGit/GTIndex.m
```diff
@@ -121,6 +121,14 @@ - (id)initWithGitIndex:(git_index *)index repository:(GTRepository *)repository
 
 #pragma mark Entries
 
+- (GTOID *)checksum {
+	const git_oid *oid = git_index_checksum(self.git_index);
+	if (oid != NULL)
```
pietbrauer: Could you add braces here?

ObjectiveGit/GTIndex.m
```diff
@@ -308,17 +316,17 @@ - (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)
 			return NO;
 		}
 
-		GTIndexEntry *blockAncestor;
+		GTIndexEntry *blockAncestor = nil;
```
pietbrauer: This is default behavior and could be omitted.

ObjectiveGit/GTIndex.m
```diff
@@ -308,17 +316,17 @@ - (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)
...
 		if (ancestor != NULL) {
 			blockAncestor = [[GTIndexEntry alloc] initWithGitIndexEntry:ancestor];
 		}
 
-		GTIndexEntry *blockOurs;
+		GTIndexEntry *blockOurs = nil;
```
pietbrauer: This is default behavior and could be omitted.

ObjectiveGit/GTIndex.m
```diff
@@ -308,17 +316,17 @@ - (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)
...
 		if (ours != NULL) {
 			blockOurs = [[GTIndexEntry alloc] initWithGitIndexEntry:ours];
 		}
 
-		GTIndexEntry *blockTheirs;
+		GTIndexEntry *blockTheirs = nil;
 		if (theirs != NULL) {
```
pietbrauer: This is default behavior and could be omitted.

ObjectiveGit/GTRepository+Status.h
```diff
@@ -57,8 +59,8 @@ extern NSString *const GTRepositoryStatusOptionsFlagsKey;
 /// An enum, for use as documented, with the `GTRepositoryStatusOptionsFlagsKey`
 /// key.
 ///
-/// See status.h for documentation of each individual flag.
-typedef enum {
+/// See status.h for documentation of each individual git_status_opt_t flag.
+typedef NS_OPTIONS(NSInteger, GTRepositoryStatusFlags) {
```
pietbrauer: Shouldn't this be `NS_ENUM`, if it was a `enum` before?

pietbrauer:
Tests for this would be really nice. @tiennou do you have any additional thoughts?

ObjectiveGit/GTRepository+Status.h
slavikus: Well, `NS_ENUM` implies a single enumeration whereas `NS_OPTIONS` allow multiple values, especially when it bridges stuff to Swift, so you can do stuff like `[ .includeUntracked, .includeIgnored ]` natively. By the meaning of the enumeration, it's really an options bitfield, hence the choice of `NS_OPTIONS` over `NS_ENUM`.

ObjectiveGit/GTIndex.m
slavikus: While I understand the compiler makes it easier by auto-nulling new declarations, I don't think an implicit `= nil` makes any harm. :)

ObjectiveGit/GTIndex.m
slavikus: Will do.

ObjectiveGit/GTIndex.m
mcyril: Slava, that's unrelated. the rationale is pretty simple (kindergarten rule) — compiler doesn't nullify local (stack allocated) scalars in C-derived languages, there is no such a thing in C as "default behaviour" which zeroes locals. explicit zeroing will prevent propagating of rubbish in case when the next condition is false. that's must have.

pietbrauer: I actually don't care too much about this, but in the [ARC transitioning release notes](https://developer.apple.com/library/content/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html) it reads "Using ARC, strong, weak, and autoreleasing stack variables are now implicitly initialized with nil.".

ObjectiveGit/GTRepository+Status.h
pietbrauer: I see, cool.

ObjectiveGit/GTIndex.m
mcyril: ok, nice, I always forget about ARC. then it could be omitted, but explicit nullify looks more semantically clean

ObjectiveGit/GTIndex.h
```diff
@@ -202,7 +207,53 @@ NS_ASSUME_NONNULL_BEGIN
 ///
 /// Returns `YES` in the event of successful enumeration or no conflicts in the
 /// index, `NO` in case of error.
-- (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)(GTIndexEntry *ancestor, GTIndexEntry *ours, GTIndexEntry *theirs, BOOL *stop))block;
+- (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)(GTIndexEntry * _Nullable ancestor, GTIndexEntry * _Nullable ours, GTIndexEntry * _Nullable theirs, BOOL *stop))block;
+
+/// An enum for use with addPathspecs:flags:error:passingTest: below
+/// See index.h for documentation of each individual git_index_add_option_t flag.
+typedef NS_OPTIONS(NSInteger, GTIndexAddOptionFlags) {
```
tiennou: Nitpick: Maybe just `GTIndexAddOptions` ? That would make for more consistent wordings (e.g. "A combination of GTIndexAddOptionFlags flags" below).

ObjectiveGit/GTIndex.h
```diff
@@ -202,7 +207,53 @@ NS_ASSUME_NONNULL_BEGIN
...
+	GTIndexAddOptionCheckPathspec = GIT_INDEX_ADD_CHECK_PATHSPEC
+};
+
+/// Add or update index entries matching files in the working directory.
+/// This method will immediately fail if the index's repo is bare.
+///
+/// pathspecs - An `NSString` array of path patterns. (E.g: *.c)
+///             If nil is passed in, all index entries will be updated.
+/// flags     - A combination of GTIndexAddOptionFlags flags
+/// block     - A block run each time a pathspec is matched; before the index is
```
tiennou: "A block that will be called on each match, before the index is changed."
"what the pathspec" => "which pathspec".
"If you pass in ~NULL~ nil"

ObjectiveGit/GTIndex.m
```diff
@@ -336,10 +344,58 @@ - (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)
 	BOOL shouldAbortImmediately;
 };
 
+- (BOOL)addPathspecs:(NSArray *)pathspecs flags:(GTIndexAddOptionFlags)flags error:(NSError **)error passingTest:(GTIndexPathspecMatchedBlock)block {
+	NSAssert(self.repository.isBare == NO, @"This method only works with non-bare repositories.");
```
tiennou: This is up for debate, but I feel like an assert is too much ? As someone who writes a GUI (where people can open mostly anything), I'd prefer to get an error back instead of `exit`.

What does `libgit2` do in that case, return `GIT_EBARE` ?

ObjectiveGit/GTIndex.m
```diff
@@ -336,10 +344,58 @@ - (BOOL)enumerateConflictedFilesWithError:(NSError **)error usingBlock:(void (^)
...
+	@onExit {
+		if (strarray.count > 0) git_strarray_free(&strarray);
+	};
+
+	struct GTIndexPathspecMatchedInfo payload = {
+		.block = block,
+		.shouldAbortImmediately = NO,
+	};
+
+	int returnCode = git_index_add_all(self.git_index, &strarray, (unsigned int)flags, (block != nil ? GTIndexPathspecMatchFound : NULL), &payload);
```
tiennou: I realize it's not from here, but `GTIndexPathspecMatchFound` sounds strange (I was wondering if it was a snuck-in flag or something).

Can you add `Callback` to the function name ?

tiennou:
Thanks for the PR, a few nitpicks but 👍 overall. Some tests would be ✨ though.

PR #624, opened 2017-03-20
GTObject method should be nullable

Uncommon:
`-[GTIndexEntry GTObject]` needs a nullable return type so that Swift sees it as a throwing function.


pietbrauer:
Cool,  thanks!

PR #628, opened 2017-05-28
Wrap git_index_add_all

alehed:
Allows the user to do something like `git add -A`

### DIFF
--- a/ObjectiveGit/GTIndex.h
+++ b/ObjectiveGit/GTIndex.h
@@ -155,6 +155,13 @@ NS_ASSUME_NONNULL_BEGIN
 /// Returns whether reading the tree was successful.
 - (BOOL)addContentsOfTree:(GTTree *)tree error:(NSError **)error;
 
+/// Add all the content of the working directory to the index. Like `git add -A`
+///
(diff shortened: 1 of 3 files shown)


alehed:
Sure, I just thought the test suite is ignored on this project.

pietbrauer:
I look at it. We "just" have one failing test on macOS which is consistent 😆. iOS is a different story.

alehed:
Yeah, would be great if all the tests worked... But I do too see the value in testing.

I'll probably get around to the tests this weekend.

pietbrauer:
Yeah, many people have spent many hours trying to fix the tests. Thanks a lot! No pressure from my side 😅

alehed:
#599 has a green iOS build, but it is already 5 months old.

pietbrauer:
Let's see what https://travis-ci.org/libgit2/objective-git/builds/237270148 brings.

pietbrauer:
Turns out it wasn't that hard. #631 has a green build. Thanks for pushing this @alehed!

alehed:
@pietbrauer Thanks for fixing the build! 👍

alehed:
I added a small test.

PR #629, opened 2017-05-28
Add merge analysis between two branches

alehed:
If you want to check the feasibility of a merge between two branches, this might come in handy.

### DIFF
--- a/ObjectiveGit/GTRepository+Merging.h
+++ b/ObjectiveGit/GTRepository+Merging.h
@@ -52,7 +52,7 @@ typedef NS_OPTIONS(NSInteger, GTMergeAnalysis) {
 /// will point to an error describing what happened).
 - (BOOL)mergeBranchIntoCurrentBranch:(GTBranch *)fromBranch withError:(NSError **)error;
 
-/// Analyze which merge to perform.
+/// Analyze which merge to perform from the branch to HEAD
 ///
 /// analysis   - The resulting analysis.
 /// fromBranch - The branch to merge from.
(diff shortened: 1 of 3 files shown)


pietbrauer (changes requested):
Thanks again! Tests would be very nice here as well.

alehed:
To my knowledge there is no way to get ObjectiveGit to give me an  "unborn" branch, that is one with an empty SHA, so GTMergeAnalysisUnborn is never returned.

alehed:
Any idea why the test fails? Locally the (iOS) tests all succeed with the newer versions of Quick and Nimble.

pietbrauer:
My guess is that we are still running tests with Xcode 8.0 on Travis until #631 is merged. There could be some changes in Xcode and/or Nimble and Quick that cause this.

alehed:
Hope the tests pass now.

tiennou:
About "unborn" : you can get one by deleting the ref pointed to by `HEAD`.

Since you can only merge a branch in `HEAD`, I'm wondering about the use case for that method.

alehed:
The reason why I needed this is for pulling a branch other than the current one. Think `git pull origin master` while you are on branch `feature-1`.

tiennou:
@pks-t @ethomson Would you guys be interested in having that capability upstream ? I'm a little wary of adding API not backed by what's in The Library...

ethomson:
Yeah; it seems that this is similar to giving `git_merge_analysis` the knowledge of a different `ours`.  I can see how that would be useful.

alehed:
Now with upstream having this feature, the PR would have to be reworked completely. I'm not working on this any more, so closing.

PR #630, opened 2017-05-28
Wrap git_merge_file

alehed:
Add a wrapper for git_merge_file

### DIFF
--- a/ObjectiveGit/GTRepository+Merging.h
+++ b/ObjectiveGit/GTRepository+Merging.h
@@ -63,7 +63,7 @@ typedef NS_OPTIONS(NSInteger, GTMergeAnalysis) {
 /// error    - The error if one occurred. Can be NULL.
 ///
 /// Returns The file content annotated with conflict markers or null on error
-- (NSString* _Nullable)stringForConflictWithAncestor:(GTIndexEntry *)ancestor ourSide:(GTIndexEntry *)ourSide theirSide:(GTIndexEntry *)theirSide withError:(NSError **)error;
+- (NSString * _Nullable)contentsOfDiffWithAncestor:(GTIndexEntry *)ancestor ourSide:(GTIndexEntry *)ourSide theirSide:(GTIndexEntry *)theirSide error:(NSError **)error;
 
 /// Analyze which merge to perform.
 ///
--- a/ObjectiveGit/GTRepository+Merging.h
+++ b/ObjectiveGit/GTRepository+Merging.h
@@ -7,6 +7,7 @@
 //
 
 #import "GTRepository.h"
+#import "GTIndexEntry.h"
 #import "git2/merge.h"
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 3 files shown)


alehed:
Just some changes I had lying around.

ObjectiveGit/GTRepository+Merging.h
```diff
@@ -52,6 +53,18 @@ typedef NS_OPTIONS(NSInteger, GTMergeAnalysis) {
...
+///
+/// The parameters taked are the ones received from `enumerateConflictedFiles`.
+///
+/// ancestor - The ancestor entry
+/// ours     - The index entry of our side
+/// theirs   - The index entry of their side
+/// error    - The error if one occurred. Can be NULL.
+///
+/// Returns The file content annotated with conflict markers or null on error
+- (NSString* _Nullable)stringForConflictWithAncestor:(GTIndexEntry *)ancestor ourSide:(GTIndexEntry *)ourSide theirSide:(GTIndexEntry *)theirSide withError:(NSError **)error;
```
tiennou: Small nitpicks :
- `*` should be right-aligned
- `withError:` => `error:`

ObjectiveGit/GTRepository+Merging.m
```diff
@@ -170,6 +172,86 @@ - (BOOL)mergeBranchIntoCurrentBranch:(GTBranch *)branch withError:(NSError **)er
...
+	if (theirData == nil) {
+		return nil;
+	}
+	NSString *theirString = [[NSString alloc] initWithData: theirData encoding:NSUTF8StringEncoding];
+	theirInput.ptr = [theirString cStringUsingEncoding:NSUTF8StringEncoding];
+	theirInput.size = theirString.length;
+
+
+	git_merge_file_result result;
+	git_merge_file(&result, &ancestorInput, &ourInput, &theirInput, nil);
```
tiennou: Missing error checking.

ObjectiveGit/GTRepository+Merging.h
tiennou: There's a few more `*` spacing weirdness in the hunk below as well.

tiennou:
Thanks for the PR ! A few more thoughts though :
- I'd say using `NSString` will prevents that from working with binary files. Would it be acceptable to switch to using `NSData` inside the method, just in case ?
- I find the `-stringForConflictWithAncestor:…` name unclear. May I suggest something like `-contentsOfDiffWithAncestor:…` or something along those lines ?

alehed:
@tiennou thanks for the feedback! I'll fix it this weekend.

alehed:
Addressed the main issues. The test is pretty messy, but I don't see a nicer or shorter way to produce conflicting commits on two different branches and merging them.

alehed:
This should be ready.

ObjectiveGit/GTRepository+Merging.m
```diff
@@ -170,6 +172,87 @@ - (BOOL)mergeBranchIntoCurrentBranch:(GTBranch *)branch withError:(NSError **)er
...
+	if (gitError != GIT_OK) {
+		if (error != NULL) *error = [NSError git_errorFor:gitError description:@"Failed to create merge file"];
+		return nil;
+	}
+
+	char *cString = malloc(result.len * sizeof(char *) + 1);
+	strncpy(cString, result.ptr, result.len);
+	cString[result.len] = '\0';
+
+	NSString *mergedContent = [[NSString alloc] initWithCString:cString encoding:NSUTF8StringEncoding];
```
tiennou: Maybe `-initWithBytesNoCopy:length:encoding:freeWhenDone:` ? That should allow you to just pass `result.ptr`/`result.len`, skipping the need for the `malloc`/`strncpy` step. You would have to `result.ptr = NULL` though, so `git_merge_file_result_free` doesn't break it by freeing before returning.

tiennou:
After a more thorough look at what `libgit2` exposes, I have a feeling that it might be a too brittle wrapper around the `git_merge_file_result` [structure](https://libgit2.github.com/libgit2/#HEAD/type/git_merge_file_result). But I'll defer to @pietbrauer on that though.

alehed:
I could wrap `git_merge_file_result` in a class if you mean that. But I don't know where else it would be used.

pietbrauer:
I'm ok with not having a wrapper for now.

tiennou:
Closing in favor of #664. Thanks for the PR @alehed, and sorry for the long turnaround!

PR #635, opened 2017-09-13
-[GTNote target] -> -[GTNote targetOID]

slavikus:
Apparently messed this one up initially: the GTNote class itself defines a read-only accessor to `targetOID`, while public header defines a `target` property. Fixing the public header so it gets proper access to `targetOID`.


pietbrauer:
Thanks!

Issue #648, opened 2018-03-19
The libssh2 story


tiennou:
I'm filing that here for ease of documentation (since macOS is obviously the biggest user of the SecureTransport/CommonCrypto "backend" combination in libgit2-land).

The reference implementation I started with is [here](https://github.com/keithduncan/libssh2/pull/2), and I have a rebased but not-working libssh2 branch of it [here](https://github.com/libssh2/libssh2/compare/master...tiennou:secure-transport).

The current problem with that implementation is that it depends on this [header](https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60118.30.2/include/CommonBigNum.h.auto.html) for its BigNum implementation, which (might) mean a definitive NACK when iOS-ing. Also, there's a vague and "concerning" comment (hence the *might*).

Additionally, libssh2 recently [changed](https://github.com/libssh2/libssh2/commit/f7daf3185a4e889fcd5272858db97764a279da1e) its crypto layer a year ago so that backends now are responsible for doing the DH computations themselves (which either requires BigNum, or could be implemented using *another* [private header](https://opensource.apple.com/source/CommonCrypto/CommonCrypto-60118.30.2/include/CommonDH.h.auto.html).

tiennou:
cc @ethomson, since we spoke about it a few weeks back.

tiennou:
AFAICS, CommonCrypto has nothing in the way of elliptic curves.

Ref: rdar://38635282, for clarification about the comment as well as usage on iOS.

tiennou:
Okay, got the backend to work. It looks like it's missing support for passworded RSA keys though…

@keithduncan Just noticed there's no licensing header in your securetransport file. Is this an oversight ?

keithduncan:
> It looks like it's missing support for passworded RSA keys though…

I can try to take a look at this if I can refresh my memory on how all this works. I wrote some backend tests in https://github.com/keithduncan/libssh2-securetransport-tests to try and exercise all the different private key encodings that are supported. Could you add a failing test there?

> Just noticed there's no licensing header in your securetransport file. Is this an oversight ?

I added a BSD license in https://github.com/keithduncan/libssh2/pull/2/commits/5341020182f63d42f57a20b8b92bb83ced5dbf49 in keeping with the project license. Thank you for letting me know :bow:

I’d love to see this backend merged upstream I just haven’t had the time to chase it down to completion 😅 let me know if I can help.

tiennou:
[...]

PR #652, opened 2018-04-09
Update openssl to 1.0.2-stable

tiennou:
@pks-t mentioned CVEs against the version we're using, so let's update.

Extracted from #645, because that fix will need more work.


tiennou:
I'd like to propose switching libssh2's backend to mbedTLS, because I don't think #648 can be realistically made to work without any external dependencies.

@pietbrauer @phatblat Opinions ?

phatblat:
I haven't heard of [mbedTLS](https://github.com/ARMmbed/mbedtls) before, but I'm game as long as there's a decent community to support it and we can build it on Apple platforms. Would this replace the CommonCrypto user by macOS in libssh2?

As for the license, they say it is [dual licensed](https://tls.mbed.org/how-to-get) as both Apache 2 or GPL 2. I would think we'd want to use Apache 2 so that it would be compatible with this repo's MIT license. I just want to make sure people can use this repo to build apps that they make money from, without having to give away the source code. It looks like they have the Apache 2 license in their github repo, so probably a non-issue.

phatblat:
What version of OpenSSL does this PR bring us up to?  Are we on 1.0.2p-dev now as the readme in the [OpenSSL_1_0_2-stable](https://github.com/openssl/openssl/tree/OpenSSL_1_0_2-stable#readme) branch says?

tiennou:
It seems the world moved since I updated my submodule, so I've pointed it at 1.0.2o which was released 27/03 instead of tracking stable.

> Would this replace the CommonCrypto used by macOS in libssh2?

We don't have that, and AFAICT we're not likely to have it (see #648). Arguably, I'm not even sure of my reasons for asking a change, since whatever we do we'd have to package ourselves, it just felt like mbedTLS might be easier but it's a hunch.

phatblat:
Ah, that’s right. Well, it would be nice to get rid of openssl

phatblat:
Can you rename the PR to reflect the version?

PR #658, opened 2018-07-20
small hack that the underlying problem is visible to the user then the error is displayed with [NSApp presentError:]

schriftgestalt:


tiennou:
It looks like something that should be handled at your level, since Objective-Git doesn't *really* expect its errors to be passed to `presentError:` (like, we're completely wrong on how error descriptions are used. This is supposed to be the alert title, so having a format-friendly version is misleading to writers and will cause strange alerts with long strings if used as-is). We also AFAIR never suggest recovery suggestions (they're not in use), only a description and failure reason (and even this is a stretch). Lastly, your addition (though it's not likely to happen) will crash while inserting a nil in a dictionary.

To be clear, I realize our error-building could be enhanced, but this feels too much like a band-aid. So, I'm 👎 on that change, though I'd still like to praise your contribution spirit.

PR #666, opened 2018-09-08
Fix bug in isValidRemoteName

alehed:
According to the docs git_remote_is_valid_name returns either 1 (success) or 0.

### DIFF
--- a/ObjectiveGit/GTRemote.m
+++ b/ObjectiveGit/GTRemote.m
@@ -104,7 +104,7 @@ NSString * const GTRemoteRenameProblematicRefSpecs = @"GTRemoteRenameProblematic
 + (BOOL)isValidRemoteName:(NSString *)name {
 	NSParameterAssert(name != nil);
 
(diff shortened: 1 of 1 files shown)


ObjectiveGit/GTRemote.m
```diff
@@ -104,7 +104,7 @@ - (NSString *)description {
 + (BOOL)isValidRemoteName:(NSString *)name {
 	NSParameterAssert(name != nil);
 
-	return git_remote_is_valid_name(name.UTF8String) == GIT_OK;
+	return git_remote_is_valid_name(name.UTF8String);
```
tiennou: I'd prefer `return (git_remote_is_valid_name(name.UTF8String) == 1 ? YES : NO);` so there's no chance of a -1 error being reported as "true", and we force-map to "BOOL"-compatible values.

alehed:
Rebased and pushed

tiennou:
Perfect, thanks !

Issue #667, opened 2018-09-27
Distributed Mac binary (via Carthage) depends on Homebrew’d openssl


mxcl:
```
$ otool -L ObjectiveGit.framework/ObjectiveGit 
ObjectiveGit:
	@rpath/ObjectiveGit.framework/ObjectiveGit (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 9.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 22.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1452.23.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1452.23.0)
```

Key lines being these:

```
/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
```

This makes it hard to distribute a binary on Mac that uses ObjectiveGit.

I realize this is probably a deep seated distribution problem, I'm willing to put in the time to fix it, but need guidance and encouragement. Thanks.

tiennou:
Hey, another GitX fork, which stumbled on the exact same problem I did just a few days ago 😉.

@mxcl Can you give a try to the PR just above, just so I have another pair of eyes ?

mxcl:
This didn't work for my GitX installation since it runs in the ObjectiveGit.framework build and at that point there is no bundle to copy the dylibs to.

[...]

PR #668, opened 2018-09-27
Automatically package OpenSSL inside the framework

tiennou:
As per the script's documentation :

> This script looks up an executable's list of shared libraries, copies
> non-standard ones (ie. anything not under /usr or /System/) into the target's
> bundle and updates the executable install_name to point to the "packaged"
> version.

In short, it will grab the Homebrew version of OpenSSL dylibs we're using and fixup the ObjectiveGit framework to use a copy of them.

It's 2018, that `install_name` should have been automated by now 😉.

Ref #667.

### DIFF
--- a/ObjectiveGitFramework.xcodeproj/project.pbxproj
+++ b/ObjectiveGitFramework.xcodeproj/project.pbxproj
@@ -1225,6 +1225,7 @@
 				79262F0F13C697BE00A4B1EA /* Copy git2 Headers */,
 				BEF7E4DF1A3A47450035BB8E /* Copy git2 Headers again */,
 				8DC2EF500486A6940098B216 /* Headers */,
+				4D751E9D215D765D003CD3CE /* Package external libraries */,
 			);
 			buildRules = (
 			);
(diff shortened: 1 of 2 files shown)


mxcl:
I had to edit the script to get it to build with Carthage.

    deps = `otool -arch x86_64 -L #{TARGET_EXECUTABLE_PATH}`

Since for some reason the `CURRENT_ARCH` variable was set to `undefined_arch`. Might be an Xcode 10 issue.

tiennou:
Okay, I've rejigged that script, because I tried to take care of codesigning, which needed refactoring, but ultimately decided that I don't care enough (also `codesign --deep` seems to be a thing). So this is the refactored version without codesigning.

I've removed the offending `CURRENT_ARCH`, since it seems to be going away, and AFAIK ~x86_64 rules the world anyways~ we only call that on a mono-arch binary.

Uncommon:
Any progress on this? It looks like the travis build failed because of a file permissions error.

Uncommon:
Thanks for the fix @tiennou. Can we please get this merged?

tiennou:
🎉

PR #695, opened 2019-06-18
Do Xcode's recommended localization naming updates

Uncommon:
Getting rid of some build warnings about localization names needing to be updated.


tiennou:
Nice, thanks @Uncommon !

PR #707, opened 2020-05-08
Pull request for deleting infoplist.strings

wilshipley:
I deleted InfoPlist.strings because it’s not used, it’s empty, and it will never be used since this is a framework not an app (and thus doesn’t need to be localized). It also caused modern versions of Xcode to compile with warnings because `English.lproj/` is deprecated (for `en.lproj` or even `Base.lproj`).


tiennou:
Thanks for the fix !

PR #712, opened 2021-04-14
Fix packaging script when paths contain spaces

lanquemar:
By default, two schemes are built: `ObjectiveGit Mac` and `ObjectiveGit iOS`.
They both contain a space character that ends up in their build directory names.

When `repackage-dylibs.rb` is run, these directories are given to `otool` and `install_name_tool` which treats spaces as argument separators.

This PR adds quotes to them.


tiennou:
Good catch, thanks !

PR #714, opened 2021-05-07
Fix warnings & update dependencies

tiennou:
Fix most warnings introduced by #713, and update dependencies to their latest published versions.


Sobhanysiamak:
Thanks for all /sms

